PR #16, opened 2017-01-10
Let executablePath default to "lean"

### LINKED ISSUE #14: Lean executable in PATH is not found
Even though I have `lean` in path, I still had to set the config option for the executable path.

Kha:
Fixes #14

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
             "properties": {
                 "lean.executablePath": {
                     "type": "string",
-                    "default": "",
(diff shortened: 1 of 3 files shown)


jroesch:
👍

leodemoura:
I'm wondering whether we need to include the `.exe` extension on Windows.

jroesch:
Most likely, I will open an issue. I would also like to support the common failure mode of putting a directory which contains Lean but is not the executable.

PR #26, opened 2017-02-10
Fix language-configuration.json

gebner:
The JSON file was not valid JSON (because of the comments), also the call to
setLanguageConfiguration was not well-typed (it expects a string, not a
DocumentFilter).


jroesch:
I fixed this many days ago by using the JSON parsing supplied by VScode itself. This seemed like the best way to do it since they will most likely invoke the same code if/when they fix the open issue.

PR #41, opened 2017-04-27
Use lean-client-js.

gebner:
This PR uses the [lean-client-js](https://github.com/gebner/lean-client-js) library to interface with the lean server.  This is the same library that I want to use for the future web editor.  @jroesch Do you agree with this change?

As a bonus, you can now restart the lean server from the toast errors.  Also restarting should work reliably now.


jroesch:
Seems like a good change to me. It generalizes a fair amount of code for all JavaScript targets.

jroesch:
I am going to use it for a bit locally, then cut a new release soon :) 👍

PR #50, opened 2017-06-11
RFC: Use previewHtml for Lean's Info View

johoelzl:
This commit changes the info view to use `previewHtml` instead of a normal text file. Currently, the only advantage is a nicer title ;-)

There is a lot still missing:
* [X] escape `<` in the Lean messages
* [X] make the file positions clickable so the user can jump to the error
* [ ] revive syntax highlighting and maybe provide additional information / feedback
* [X] uses now a editor context menu, maybe switch to title menu
* [X] if VSCode's HTML preview listens to changes from the `TextDocumentContentProvider` we can provide more fine grained updates for errors and the goal view. I marked this as finished. It seams to be fast enough and we only update messages when we actually have new data and not a new position.

@jroesch @gebner what do you think? Is this the right way to go?


gebner:
> Is this the right way to go?

Yes, this is exactly how I imagined it.  Thanks for taking up the initiative!

> escape < in the Lean messages

Ah, little Bobby Linear Order...

> make the file positions clickable so the user can jump to the error

:+1: 

The built-in markdown preview automatically scrolls the preview when you move around in the editor.  Should we do that as well?

> make {update paused | actively updating} and {showing|hiding} error messages active,

Just in case you didn't already know.  This is very easy to do with an `<a href=...>` tag: https://code.visualstudio.com/docs/extensionAPI/vscode-api-commands#_links

> if VSCode's HTML preview listens to changes from the TextDocumentContentProvider

I would have expected this to *just work*.  Did you encounter any problems where the change events from the InfoProvider were ignored?

README.md
```diff
@@ -41,6 +41,14 @@ This extension contributes the following settings:
 * By default, vscode will complete `then` to `has_bind.and_then` when you press enter.  To disable this behavior, set `"editor.acceptSuggestionOnEnter": false`
 * If you like colored brackets, try out [Bracket Pair Colorizer](https://marketplace.visualstudio.com/items?itemName=CoenraadS.bracket-pair-colorizer).
 
+## Development
```
gebner: This section is a really good idea.

src/extension.ts
```diff
@@ -135,11 +135,12 @@ export function activate(context: vscode.ExtensionContext) {
...
         infoProvider,
         vscode.workspace.registerTextDocumentContentProvider(
-            infoProvider.scheme, infoProvider),
+            infoProvider.leanGoalsUri.scheme, infoProvider),
         vscode.commands.registerTextEditorCommand('lean.infoView', (editor) => {
-            vscode.workspace.openTextDocument(infoProvider.leanInfoUrl)
-                .then(doc => vscode.window.showTextDocument(
-                    doc, editor.viewColumn + 1));
+            let column = editor.viewColumn + 1;
+            if (column = 4) column = vscode.ViewColumn.Three;
```
gebner: I'm pretty sure the `column = 4` condition has a side effect.

johoelzl: Haha, ouch. I'm too used to theorem provers.

johoelzl:
> The built-in markdown preview automatically scrolls the preview when you move around in the editor. Should we do that as well?

This is a good idea! Something I want to try. I guess this will take some experimentation to avoid an annoying behaviour where the user looses track what the current state is. Also I would like to avoid too much movement in the goal view when the user goes one line up or down in the editor.

>> if VSCode's HTML preview listens to changes from the TextDocumentContentProvider

> I would have expected this to just work. Did you encounter any problems where the change events from the InfoProvider were ignored?

I wasn't clear here. My idea would be that the HTML page behind `lean-info:goal-view` would be just:
```
<iframe href="lean-info:current-goal"></iframe>
<iframe href="lean-info:msg/12345678"></iframe>
<iframe href="lean-info:msg/C0FEBABE"></iframe>
<iframe href="lean-info:msg/C0FEFE00"></iframe>
```
And when the user moves the cursor around we only signal an update to `lean-info:current-goal`.

But this is something we can think about when the HTML rendering is too slow.

src/infoview.ts
```diff
@@ -36,26 +38,34 @@ class InfoDocument implements Disposable {
...
-        this.renderGoal();
-        this.renderMessages();
+        this.contents = this.renderHeader() + this.renderGoal() + this.renderMessages();
+    }
+
+    private stylesheet() {
+        return `#settings { display: block; text-align: right; }
+            h1 {
+                display: block;
+                background-color: blue;
```
gebner: I think we need a different stylesheet depending on whether we have a light or a dark background:

info-view-html-bg

johoelzl: Absolutely :-) Is there an easy access to the style colors? Or should we just provide a light and a dark version? (like the progress images)

johoelzl: VS Code sets the theme bodys class e.g. `<body class="vscode-light">`. So we can easily setup a CSS supporting different themes.

jroesch: I think we can get all the styling information from the editor itself, it would be good if we could use a style consistent with the user's preferences.

johoelzl: I'm currently playing around with having a more separate view on the goal and the list of error messages. I guess I will introduce a split view, with a fixed (user setable) size for the goal view.

@jroesch I mostly try to stick to this, but I'm not sure how we can reuse existing theme colors. For example I want to have different colors for the message titles, depending if they are errors, warnings or informations.

gebner: The split view sounds reasonable.

Re: color themes.  I absolutely agree that we should try to fit in with the vscode theme.  Unfortunately, I don't think there is any way to obtain theme colors from an extension.  The HTML preview gets injected with a few values, but that's just [foreground/background colors, and fonts](https://github.com/Microsoft/vscode/blob/3b8a3ca25e5ed631c6a786792ac85cc817b09891/src/vs/workbench/parts/html/browser/webview.ts#L182-L186).

I think the best that we can do is pick colors that fit with the default Dark+ and Light+ themes.  The info view should really look like a part of vscode, and not call attention to itself.

johoelzl:
This is now a complete change of the `infoview`: it only shows the tactic state + errors at the current position. I want to extend this, so that if no error is available we show the previous ones (with an indication that they are in front of the current cursor position).

The idea is that if people want to see all errors they can use the problem view. Or maybe have two different modes for the info view: current tactic state + current errors  OR   all errors with text editor tracking.

.vscode/launch.json
```diff
@@ -7,7 +7,7 @@
             "type": "extensionHost",
             "request": "launch",
             "runtimeExecutable": "${execPath}",
-            "args": ["--extensionDevelopmentPath=${workspaceRoot}" ],
+            "args": ["--extensionDevelopmentPath=${workspaceRoot}", "/home/jholzl/Projects/lean-scratch/coinduction.lean"],
```
gebner: You accidentally committed a local path here.  I have never needed to add anything here, F5 always re-opens the last folder (that I opened in F5).

src/infoview.ts
```diff
@@ -5,230 +5,212 @@ import {
...
-        this.render();
-        this.changedEmitter.fire();
-    }
+    private render() {
+        return `<!DOCTYPE html>
+            <html>
+            <head> 
+            <meta http-equiv="Content-type" content="text/html;charset=UTF-8">
+            <style>
+            div
```
gebner: Is there a reason you put the stylesheet back into the `infoview.ts` file?  With an extra file, there's at least syntax highlighting for the CSS.

gebner:
> it only shows the tactic state + errors at the current position.

I just tried it and find it very uncomfortable to use.  The cursor needs to be at exactly the right position to see the message, and this position is not even necessarily the begin of the squiggles...  Can we at least show all messages for the current line?

> they can use the problem view

The problem view is next to useless for looking at errors, since it squashes everything into a single line and doesn't have syntax highlighting.  Even just using it to select the errors is cumbersome, since you need to figure out which line in the problems view corresponds to the message in the editor window.

gebner:
BTW, the stylesheet is progressing nicely.  I also agree that it is much less crowded if not all messages are shown at the same time.

I think we should try to use the editor font settings for the info view, i.e., use `workspace.getConfiguration('editor').get('fontFamily')` and `workspace.getConfiguration('editor').get('fontSize')`.

src/infoview.ts
```diff
@@ -1,185 +1,216 @@
...
+            if (i == -1) {
+                return `<div class="message ${m.severity}"><h1>${m.severity}: ${m.caption}</h1><pre>${m.text}</pre></div>`
             }
-            const doc = new InfoDocument(this.server, this.leanDocs);
-            doc.onChanged(() => this.changedEmitter.fire(uri));
-            this.documents.set(uriString, doc);
-        }
-        return this.documents.get(uriString).text;
+            return `<div class="message ${m.severity}">
+                <h1>${m.severity}: ${m.caption} ${m.text.substring(0, i)}</h1>
```
gebner: The `m.text.substring` causes the first line of the trace output to be displayed as part of the caption.  Is there any situation where we might actually want that?

gebner:
> I think we should try to use the editor font settings for the info view

At least with my settings, this makes a huge difference.  Otherwise Electron uses the default monospace font, which doesn't have good support for Unicode symbols and also looks very out of place next to the editor font.

```diff
diff --git a/src/infoview.ts b/src/infoview.ts
index f35a878..df4bb4b 100644
--- a/src/infoview.ts
+++ b/src/infoview.ts
@@ -145,6 +145,11 @@ export class InfoProvider implements TextDocumentContentProvider, Disposable {
   padding: 0px;
 }
 
+pre {
+    font-family: ${workspace.getConfiguration('editor').get('fontFamily')};
+    font-size: ${workspace.getConfiguration('editor').get('fontSize')}px;
+}
+
 h1
 {
   display: block;
```

src/infoview.ts
johoelzl: When I used a `<link src="stylesheet.css" ...>` the CSS sporadically wasn't applied when waking over the text. Embedding it into the HTML helped fixing this. But you are right, the extra file was very helpful for me. I will try to load it on plugin start, and add a configuration option to add extra CSS snippets.

src/infoview.ts
johoelzl: Currently the first line is colored, e.g. for error it is red. So it draws a lot of attention, also `m.caption` is usually empty, here the first line takes the role of `m.caption`. Maybe this is something which should change in Lean?

johoelzl:
Using `editor.fontFamliy`/`editor.fontSize`: Will add this. Then, in my current state the HTML view does not add much ;-)

W.r.t. messages / goal state: I will readd the message view and add a _editor title (context) menu_ to switch between messages, and the current goal view, and maybe a  combined view.

.vscode/launch.json
johoelzl: Ah sorry. While it opens the last folder, I had the problem that it also opened a new file. And not the file I used to test errors.

src/infoview.ts
gebner: > So it draws a lot of attention

I'm not sure, the light red text of the caption seems a bit lighter than the black text; at least with my settings:
vscode-faint-red
It also puts the first line away from the main text, indented, and in a different color, which is something we don't do anywhere else.  It actually makes it kind of hard to read for me, but maybe I'd get used to it.

What I'm mainly concerned about is when text gets split between the caption and the pre block:
vscode-split-text

> Maybe this is something which should change in Lean?

I have actually thought about the opposite: removing captions altogether.  We don't show them in emacs, or in the vscode diagnostics.

Personally, I think that we should not try to have special captions, or even guess that the first line could be used as a caption.  If we really want to highlight portions of a message, then we should use structured markup (server-side and also transmitted over the server interface): this would also allow us to tag the parts of a message which should be syntax highlighted as lean code.

For now, I'd strongly prefer that we do not try to guess any captions--this breaks very easily.  We can add more specific captions on the server side, if you think that's a good idea.

gebner:
Another comment on the colors: we need to use a different color for warnings on light background.  The yellow is almost unreadable on white background.

johoelzl:
So the message list is back, now with better revealing of the errors at the current cursor position.

I didn't yet implement better heuristics for the errors in the goal display.
One option is surely to use the ones from the current line. But maybe only the last ones one the same line before the current position?

johoelzl:
Ah, I didn't realize that there are errors at the end of the line. A concrete example where there is too much information on the line is something like: `⟨_, _, _, _⟩`. I write this kind of terms for proofs of a set of conjunctions and existentials.  This often happens in big contexts, i.e. where each shown placeholder has a lot of local variables.

But as I write this, I realized that the messages for errors on the right are not the big problem, as they are after the errors at the current position. So, what do you think about the following heuristic:
1) go to the left until you find a error, starting from the cursor
2) show all errors on this line from this point onward
My rational is that I want to avoid using the mouse to go to the infoview and scroll around, concentrating on the cursor position in the line should reduce this requirement. I try to implement this and see how natural it feels.

johoelzl:
I implemented the previous move mentioned heuristic, and also when hovering over a message, it is indicated in a visible editor

src/infoview.ts
```diff
@@ -167,13 +167,32 @@ export class InfoProvider implements TextDocumentContentProvider, Disposable {
...
-                    m.pos_col == this.curPosition.character);
+                    m.pos_line == this.curPosition.line + 1)
+                .sort((a, b) => a.pos_col - b.pos_col);
+            let startColumn = undefined;
+            let startPos = null;
+            for (let i = 0; i < msgs.length; i++) {
+                if (this.curPosition.character < msgs[i].pos_col) break;
+                if (this.curPosition.character == msgs[i].pos_col) {
+                    startColumn = this.curPosition.character;
+                    startPos = i;
```
gebner: This doesn't work if there are multiple messages at the same position, right?

johoelzl: `startPos` should be the **first** message with column `curPosition`. Note the `break` afterwards.

gebner: Ok, reasoning about imperative programs is not my strong side. :smile:

gebner:
I just tried your latest commit.  All messages (not just from the current line) are shown at once, and I don't see goals.  Am I doing something wrong?

johoelzl:
If you go to the preview editor, you can select in the editor's title context menu (the three dots) a menu point to toggle the display mode. I will try to make this more visible (I think we can add a button, maybe we also want to have a key-code)

gebner:
Ok, got it, everything works now.  I'm not sure I would have ever found it.  I kind of liked the links you had at the top right.

BTW, I get a vertical scrollbar now.  Also when I resize the preview window, then it flickers pretty heavily.

gebner:
The "everything right of the next error to the left" heuristic works for me.

The hover is awesome!

The marker span doesn't move when I scroll the info view.

johoelzl:
The vertical scroll bar was just a CSS fix. I have now idea why the preview windows flickers, could be a problem of the preview html view itself (I will compare it to the markdown preview). The marker is now attached to the right element and should follow when scrolling.

I added icons to select start/stop and toggle mode. The problem is that these icons are only shown when the preview window is focused, I'm not sure how good this discoverabiltiy is. The commands are also seen in the command palette.

src/infoview.ts
```diff
@@ -65,30 +66,31 @@ export class InfoProvider implements TextDocumentContentProvider, Disposable {
...
+                this.setMode(DisplayMode.AllMessage);
+                this.openPreview(editor);
             }),
-            commands.registerTextEditorCommand('lean.infoView', (editor) => {
-                let column = editor.viewColumn + 1;
-                if (column == 4) column = ViewColumn.Three;
-                commands.executeCommand('vscode.previewHtml', this.leanGoalsUri, column,
-                    "Lean Messages");
+            commands.registerTextEditorCommand('lean.stopUpdating', (editor) => {
+                this.stopUpdating();
```
gebner: Should we make this a toggle?  Right now the button is pretty useless after you've pressed it once.

package.json
```diff
@@ -185,24 +189,19 @@
             ],
             "editor/title": [
                 {
-                    "command": "lean.infoView.toggleMode",
-                    "when": "resourceScheme == lean-info",
+                    "command": "lean.displayList",
+                    "when": "editorLangId == lean",
```
gebner: Can we show the buttons on top of the info view as well?  At the moment they vanish as soon as soon as you focus the info view (e.g. if you select something).

package.json
```diff
@@ -162,6 +160,12 @@
                 "when": "editorTextFocus && editorLangId == lean"
             },
             {
+                "command": "lean.displayList",
+                "key": "ctrl+shift+alt+enter",
```
johoelzl: `ctrl+shift+enter` is already emacs :-) I generally like to use `alt`-versions, it feels natural. And in this case most keyboard have `ctrl+shift+alt` easily accessible with one hand.

Another idea would be to go full emacs and use a prefix like `ctrl+L` and then have `ctrl+L g`, `ctrl+L l`, `ctrl+L s` or similar for Lean interactions.

package.json
johoelzl: Okay, I will add this.

package.json
johoelzl: Also I'm open to other suggestions for keys/icons/css etc...

gebner: I don't think these key combinations are a big issue.  The most common workflow is probably to open the info view immediately after starting vscode and then just watching the updates (at least that's what I do).  In this scenario, the keybinding is only used once per session and should not cause RSI.

src/infoview.ts
johoelzl: I don't like this kind of toggles in the UI. In this case we do not even get an immediate feedback at the button (I'm not sure if and how we can change the icon), only at the text which is a little bit away. Especially something like a _save current information_, when the user accidentally clicks the stop button and didn't realize that the `stopped updating` also means that the stop button is now a continue button.

johoelzl:
@gebner I think this PR is finished. In the next days I will do some Lean proofs and see how well the new info view works.

What is your opinion on the stop button? If you want I can add two buttons:  `⯈  ||` in the info view itself, but this adds another interaction between the HTML side and our extension.

Anything you want to see changed, anything I forgot?

src/infoview.ts
gebner: I see what you mean.

> If you want I can add two buttons: ⯈ || in the info view itself

I think that would be best.

gebner:
> I think this PR is finished.

I agree.  I am tired of switching branches just to have the info view goodness, so it's in master now.

> Anything you want to see changed, anything I forgot?

* Maybe remove the underlines on the links (captions), we already have a large line there anyways.
* The pause button looks like a door, but I don't think I could do it better.

I think I've already complained enough last week.  Let's see what other people think.

johoelzl:
@gebner Thanks for the valuable and constructive feedback!

PR #51, opened 2017-06-12
fix(package.json): only show commands acting on Lean files when a Lea…

### LINKED ISSUE #49: confusing response to check open files
screen shot 2017-06-10 at 11 39 46 am

maybe i'm missing something obvious, though its definetly true that the the highlighting ux for errors is pretty nice

johoelzl:
…n file is open. Also activate plugin on all other commands

This should fix #49

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,62 +51,74 @@
         "commands": [
             {
                 "command": "lean.restartServer",
-                "title": "Lean: Restart",
+                "category": "Lean",
+                "title": "Restart",
                 "description": "Restart the Language Server"
             },
             {
                 "command": "lean.displayGoal",
-                "title": "Lean: Display Goal",
+                "category": "Lean",
+                "title": "Display Goal",
                 "description": "Display the goal at the current position."
             },
             {
                 "command": "lean.infoView",
-                "title": "Lean: Open Info View",
+                "category": "Lean",
+                "title": "Open Info View",
                 "description": "Displays the goal and error messages in a separate window"
             },
             {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -136,19 +148,43 @@
...
                 "command": "lean.batchExecute",
                 "key": "ctrl+shift+r",
                 "mac": "cmd+shift+r",
-                "when": "editorTextFocus"
+                "when": "editorTextFocus && editorLangId == lean"
             }
-        ]
+        ],
+        "menus": {
+            "commandPalette": [
```
gebner: What effect does this section have?  I can already access all commands from the command palette (the one you can reach with `ctrl+p >` or `ctrl+shift+p`).  Is there another command palette somewhere?

package.json
```diff
@@ -51,62 +51,74 @@
         "commands": [
             {
                 "command": "lean.restartServer",
-                "title": "Lean: Restart",
+                "category": "Lean",
```
gebner: Does the category show up anywhere?

johoelzl: The category is used by the `commandPalette` menu to construct the name.

package.json
johoelzl: Yes, `commandPalette` (`ctrl+shift+p`) is usually filled up with contributions in `commands`. By adding our own entries to `commandPalette` we can overwrite the `when` attribute. So we can control that some commands are only shown in the command palette when we are in a Lean editor. So when you are in a normal `*.txt` file, `lean.showGoals` should not be available in `ctrl+shift+p`.

PR #53, opened 2017-06-18
Set source in diagnostics: goto erro shows multiline errors

johoelzl:
I'm not sure if this is a bug in vscode, but attaching a source to the error
message changes the inline diagnostic message to show more than just the first
line, see https://github.com/Microsoft/vscode/blob/master/src/vs/editor/contrib/gotoError/browser/gotoError.ts#L204

With this `F8` and `Shift-F8` show now a multi-line error.

correct go to error


gebner:
Bug in vscode or not, this is a non-negotiable feature. :smile: 

Feel free to report a bug with vscode, though.  This really looks like an oversight to me.

PR #54, opened 2017-06-19
Add explicit pause/continue button in infoview; remove underlines from links

johoelzl:
Two small changes:
* the stop button in the editor header moved to the infoview and toggles between pause/continue; this time the state change is visible
* remove underlines from `<a>` elements


leodemoura:
I think I'm going to switch to VS-Code :) 
The Lean plugin is getting nicer and nicer.

PR #55, opened 2017-06-20
infoview: mark current messages (to be used by CSS); reload css on configuration change

johoelzl:
With these changes users can experiment with css:

```json
"lean.infoViewStyle": "div.message { text-color: #888; } .marked { text-color: red }"
```


gebner:
Marked means the message with the triangle on the left, right?  So this is mainly useful when showing all messages?

johoelzl:
Yes. For the goal state view we do not have any marked messages. IN CSS we can select the messages list by `body[attr=data-messages]`

`next-marked` is the class if the cursor is on the way to the next message, but not there yet.

PR #74, opened 2018-06-03
adding unicode Mathematical Alphanumeric Symbols

semorrison:
In the `mathcal` and `mathfrak` alphabets in unicode, some letters appear in different code blocks, and these were omitted from `translations.json`.

I've also put in all the "blackboard bold" letters uniformly, but they have awkward names for now. My suggestion would be to have these uniformly as, e.g. `\bbA`, `\bbB`, and remove the existing `\bc`, etc., but this might annoy people.


gebner:
Looks good to me.  Feel free to add `\bbA`, etc. as well, they don't break anything as far as I can tell.

gebner:
I don't think we should remove `\bc` though, somebody might be using that.

semorrison:
Is it okay if I remove the current "Bbb{N}" and friends, which seem pretty
unusable?

gebner:
> Is it okay if I remove the current "Bbb{N}" and friends, which seem pretty unusable?

These abbreviations have originally been exported from the emacs lean-mode, and most of these them are from the quail package.  See [here for the abbreviations that we added in emacs](https://github.com/leanprover/lean-mode/blob/cd3374cc548106472b38484a7d0046bd3c95695e/lean-input.el#L196).

I don't think anybody uses `\Bbb{N}`, so I'm fine with removing it.

gebner:
@Kha You may want to copy the new abbreviations to lean-mode.

PR #87, opened 2018-09-04
fix two syntax highlighting examples from zulip

bryangingechen:
This PR fixes the following:
```lean
variables (A : Type*)
class MWE extends has_zero A
```
```lean
set_option class.instance_max_depth 5
```
[Reported here](https://leanprover.zulipchat.com/#narrow/stream/113488-general/subject/VScode.20extension/near/133300183).

The change addressing the second example could be applied more generally, but I haven't bothered to do this since I don't yet see any other _valid_ lean syntax with incorrect highlighting.

Edit: another similar example fixed:
```lean
#check exists.intro
#check exists.elim
```

### DIFF
--- a/syntaxes/lean.json
+++ b/syntaxes/lean.json
@@ -14,7 +14,7 @@
         {"include": "#definitionName"},
         {"match": ","}
       ],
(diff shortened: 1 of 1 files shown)


gebner:
Thanks!  Can you resolve the merge conflict?

PR #88, opened 2018-09-09
fix comment highlighting bug: -- disabling -/

bryangingechen:
This PR is aimed at fixing examples like this:
```lean
/- -- -/
#print true
```
The extension currently highlights everything as a comment, since the `--` marker prevents the `-/` from being detected.

**edit:** the first commit wasn't quite right; [the scanner](https://github.com/leanprover/lean/blob/ceacfa7445953cbc8860ddabc55407430a9ca5c3/src/frontends/lean/scanner.cpp#L341) only pays attention to the nesting of block comments and not doc comments. The second commit thus also fixes the highlighting of:
```lean
/---/
example : true := trivial
```
```lean
/--/--/
example : true := trivial
```
```lean
/--/- /--/
example : true := trivial
```
etc.

**edit 2:** the third commit now treats the "mod doc block" `/-!` syntax. I also reordered the include rules in the repository to better mirror the structure of the scanner, which slightly simplified things.

### DIFF
--- a/syntaxes/lean.json
+++ b/syntaxes/lean.json
@@ -55,23 +55,27 @@
     { "match": "\\b([0-9]+|0([xX][0-9a-fA-F]+))\\b", "name": "constant.numeric.lean" }
   ],
   "repository": {
-    "blockComment": {
-      "begin": "/-[^-]", "end": "-/", "name": "comment.block.lean",
-      "patterns": [{"include": "#blockComment"}]
-    },
-    "docComment": {
-      "begin": "/--", "end": "-/", "name": "comment.block.documentation.lean"
-    },
     "dashComment": {
       "begin": "(--)", "end": "$",
       "beginCaptures": {"0": {"name": "punctuation.definition.comment.lean"}},
       "name": "comment.line.double-dash.lean"
     },
+    "docComment": {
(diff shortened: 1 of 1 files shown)


syntaxes/lean.json
```diff
@@ -55,23 +55,27 @@
...
     "dashComment": {
       "begin": "(--)", "end": "$",
       "beginCaptures": {"0": {"name": "punctuation.definition.comment.lean"}},
       "name": "comment.line.double-dash.lean"
     },
+    "docComment": {
+      "begin": "/--", "end": "-/", "name": "comment.block.documentation.lean"
+    },
+    "modDocComment": {
+      "begin": "/-!", "end": "-/", "name": "comment.block.documentation.lean"
```
bryangingechen: I've left the "name" as "comment.block.documentation.lean" for now as I haven't yet figured out what the difference between "mod doc" and "doc" block comments yet.

gebner: Module documentation comments come at the beginning of a file.  I don't think they are used anywhere at all.

bryangingechen: Ah, so "mod" stands for "module"! Makes sense.

Issue #100, opened 2018-12-07
option to supress variables bound to _, in info view


ratmice:
When binding like: (\lam a _, a), in the info view, it will list a variable e.g. _x : ...
It would perhaps be a nice enhancement if it were possible show/hide these bindings beginning with _ in the info view.

bryangingechen:
I've just made PR #101 which is an attempt in this direction. Please test it out and leave comments there!

ratmice:
With Bryan's PR merged, i think this can be closed.

PR #101, opened 2018-12-27
tactic state filter [RFC]

bryangingechen:
This PR addresses issue #100. It adds a dropdown menu to the info view which allows filtering of the tactic state with regex strings (the list of options is configurable via editing the new settings `lean.infoViewTacticStateFilters` and `lean.infoViewFilterIndex`).

### DIFF
--- a/media/infoview.css
+++ b/media/infoview.css
@@ -39,6 +39,16 @@ a {
   text-decoration: none;
 }
 
+div#filter {
+  display: block;
+  position: absolute;
+  top : 0px;
+  right: 90px;
+  width: auto;
+  padding: 3px 0px 1px 0px;
+  z-index: 2;
+}
+
 div#run-state {
   display: block;
   position: fixed;
--- a/package.json
+++ b/package.json
@@ -94,6 +94,20 @@
 					"type": "string",
 					"default": "",
 					"description": "Add an additional CSS snippet to the info view."
+				},
+				"lean.infoViewTacticStateFilters": {
+					"type": "array",
+					"default": ["^(?!_)."],
+					"description": "An array of regular expression strings that can be used to filter the tactic state in the info view. When a filter is enabled, only lines matching the regex are displayed. Set to an empty array '[]' to hide the filter select dropdown.",
+					"items": {
+						"type": "string",
+						"description": "a properly-escaped regex string, e.g. '^(?!_).' will match all lines in the tactic state that do not begin with an underscore."
+					}
+				},
+				"lean.infoViewFilterIndex": {
+					"type": "number",
+					"default": -1,
+					"description": "Index of the filter applied to the tactic state (in the array infoViewTacticStateFilters). An index of -1 means no filter is applied."
 				}
 			}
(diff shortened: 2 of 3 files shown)


PatrickMassot:
Wonderful! Two comments: 
* From a visual perspective, the filter drop menu is really strong, at least on a dark background. Do you think it could be less visible?
* Wouldn't be simpler to list regex matching lines you want to exclude? I think this feature is there to remove stuff we don't want to see, not to find stuff we are looking for.

PatrickMassot:
I'm away from Lean until January 1st, but what you wrote sounds great.

ratmice:
I tried it out, i think it is great!

There is one thing I thought might be worth considering (since it would seem to mean a change in preferences), but overall it's not difficult to just | regex's together that I think it's necessary.
would be for the preferences, something to the effect of:
// takes the place of "match":true
"lean.infoViewTacticStateFiltersInclude": [],
// takes the place of "match":false
"lean.infoViewTacticStateFiltersExclude": [],
// allows an array of selectively enabled filters to match
"lean.infoViewFilterIndexes": {"include":[-1],"exclude":[-1]}

The primary difference being an array of ints, rather than an int, allowing you to selectively enable multiple filters.
I personally don't see myself having enough filters that having e.g. 2 filters, and a 3rd regex which or's them is going to be a problem.  But it seemed worth considering while the preferences are not yet fixed.

bryangingechen:
It's definitely worth thinking more now about what an ideal filtering feature would be like. One thing that comes to mind is functionality which would let you hide specific items from view interactively, by e.g. clicking on an x next to them in the tactic state display. Having settings like what you described might make that easier, but I'd need to think about / get more feedback on what a good workflow would be.

Here's a proposal:
* By clicking on the tactic state, the user can add and apply new exclusion filters.
* The user can select previously saved filters to apply using some kind of checkbox UI, e.g. VS Code's built-in ["QuickPick"](https://code.visualstudio.com/updates/v1_22#_multiselect-quickpick). This might replace the current dropdown menu, but I'm not sure.
* Maybe there should be a command to save a filter selection as a new regex object?
* (As a reminder to myself) there should be some indication on the info view when filters are active, maybe keep the dropdown around for this?
* For deleting and more complicated edits of filters, I'm still leaning towards making the user edit the settings by hand though. (For that reason the filter object representation should remain simple and intuitive).

I'd be curious to hear more thoughts and suggestions. Also, what sorts of default filters would be useful?

gebner:
One thing that really confused me about the UI is that the filter dropdown is always visible, even if there is no goal.  It is also not obvious that the filter only applies to the goal (shouldn't it also apply to the error messages as well?).

For example if there are only error messages, then it appears over the heading of the first error message and it looks like you could filter the message.

gebner:
Please also run `npm i` to format the `package.json` file.

bryangingechen:
I decided to just make the changes suggested by @gebner for now. The more ambitious filter idea can wait until there's more demand / feedback.

PR #103, opened 2018-12-28
fix hover functionality in info view

bryangingechen:
Depends on #101.

The code for hovering in the info view was written for the old `previewHTML()` command. I've moved the code from infoview-ctrl.js to infoview.ts so that it can make use of the new `acquireVsCodeApi()` object for passing messages to the extension.

This PR also contains a fix so that clicking on the message titles in the info view now scrolls you to the correct place in the open editor (rather than the current behavior of opening a new editor window).

### DIFF
--- a/src/infoview.ts
+++ b/src/infoview.ts
@@ -69,8 +69,6 @@ export class InfoProvider implements Disposable {
                 this.rerender();
             }),
             commands.registerCommand('_lean.revealPosition', this.revealEditorPosition),
-            commands.registerCommand('_lean.hoverPosition', (u, l, c) => { this.hoverEditorPosition(u, l, c); }),
-            commands.registerCommand('_lean.stopHover', () => { this.stopHover(); }),
             commands.registerCommand('_lean.infoView.pause', () => {
                 this.stopUpdating();
             }),
@@ -138,6 +136,17 @@ export class InfoProvider implements Disposable {
                 });
             this.webviewPanel.webview.html = this.render();
             this.webviewPanel.onDidDispose(() => this.webviewPanel = null);
+            this.webviewPanel.webview.onDidReceiveMessage((message) => {
+                switch (message.command) {
+                    case 'hoverPosition':
+                        this.hoverEditorPosition(message.data[0], message.data[1], message.data[2],
+                            message.data[3], message.data[4]);
+                        return;
+                    case 'stopHover':
+                        this.stopHover();
+                        return;
+                }
+            }, undefined, this.subscriptions);
         }
     }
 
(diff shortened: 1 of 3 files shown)


PatrickMassot:
Looks like thid has more commits than intended...

bryangingechen:
Indeed, this builds on some functionality from #100. I might have been able to make it completely independent with some git-fu but it might be easier just to look at this after the other one is merged.

src/infoview.ts
```diff
@@ -416,12 +446,15 @@ export class InfoProvider implements Disposable {
         return this.curMessages.map((m) => {
             const f = escapeHtml(m.file_name); const b = escapeHtml(basename(m.file_name));
             const l = m.pos_line.toString(); const c = m.pos_col.toString();
+            const el = m.end_pos_line ? m.end_pos_line.toString() : l;
```
gebner: `m.end_pos_line || l`.  I don't think we need to call `.toString()` here.

src/infoview.ts
```diff
@@ -351,6 +360,27 @@ export class InfoProvider implements Disposable {
                     function selectFilter(value) {
                         vscode.postMessage({command: 'selectFilter', filterId: parseInt(value)});
                     }
+                    function setupHover() {
```
gebner: Please move this function back to `infoview-ctrl.js`.

src/infoview.ts
```diff
@@ -69,8 +69,8 @@ export class InfoProvider implements Disposable {
                 this.rerender();
             }),
             commands.registerCommand('_lean.revealPosition', this.revealEditorPosition),
-            commands.registerCommand('_lean.hoverPosition', (u, l, c) => { this.hoverEditorPosition(u, l, c); }),
-            commands.registerCommand('_lean.stopHover', () => { this.stopHover(); }),
+            // commands.registerCommand('_lean.hoverPosition', (u, l, c) => { this.hoverEditorPosition(u, l, c); }),
```
gebner: If you remove the `_lean.hoverPosition` command, then please delete it completely.  Do not just comment it out.

src/infoview.ts
```diff
@@ -188,18 +195,20 @@ export class InfoProvider implements Disposable {
         for (const editor of window.visibleTextEditors) {
             if (editor.document.uri.toString() === uri.toString()) {
                 const pos = new Position(line - 1, column);
-                window.showTextDocument(editor.document);
```
gebner: What's the rationale for this change here?  Again, please do not comment out, just remove it.

gebner: Ok, then please remove the line.

src/infoview.ts
bryangingechen: I just tried this and I'm having some trouble with tslint. How do I get typescript to recognize `acquireVsCodeApi`?

bryangingechen:
I'm going to try to clean this up so that this can be merged before the filter branch.

src/infoview.ts
gebner: The quick fix is to remove `@ts-check` from that file.  Ideally we'd use the typescript compiler to compile that file as well (then we could use `import`, `declare function`, etc.) but I'm not sure how to set that up.

bryangingechen:
Done.

Issue #104, opened 2019-01-08
Automatically show info view when opening a Lean file


gebner:
As suggested at the Lean Together 2019 workshop.

PR #105, opened 2019-01-09
auto-open info view

bryangingechen:
Addresses #104

One thought: this could potentially be annoying for anyone who is just using this extension for unicode input for other languages. However, it seems to me we're already automatically loading a bunch of other Lean-only stuff when the extension is activated, so that should be addressed at a higher level, if at all.  In any case the auto-open can be turned off with the config option `lean.infoViewAutoOpen` if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ We currently support a variety of features.
 * Auto-completion based on context and type via the Lean server
 * Error messages / diagnostics
 * Customizable Unicode input support (e.g. type `\la`+<kbd>tab</kbd> to input `λ`)
-* Info view window to show goal, tactic state, and error messages (ctrl+shift+enter for local goal view or ctrl+shift+alt+enter for all messages)
+* Info view window to show goal, tactic state, and error messages (<kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>enter</kbd> for local goal view or <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>alt</kbd>+<kbd>enter</kbd> for all messages)
 * Batch file execution
 * Search for declarations in open files (<kbd>ctrl</kbd>+<kbd>p</kbd> #)
 * Region of interest checking (i.e. control how much of the project is checked automatically by Lean)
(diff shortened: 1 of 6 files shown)


bryangingechen:
I'm going to see if I can get the extension to automatically restart the lean server when it is activated.

gebner:
> One thought: this could potentially be annoying for anyone who is just using this extension for unicode input for other languages.

This is a dealbraker.  The extension is activated whenever the users opens either a markdown or a Lean file (and maybe more in the future).  We can't just open windows on activation.

> However, it seems to me we're already automatically loading a bunch of other Lean-only stuff when the extension is activated, [...]

Just registering commands, that's fine.  We don't even start the server if we don't see a `.lean` file.

I'd suggest to listen to the `server.restarted` event, it is fired every time the server is (re-)started.  I think we should just open the info view window the first time that event is fired.

bryangingechen:
I've noticed that when developing the extension I always have to restart the server manually when I launch the debugger for things to work. Is this indicative of something we should address by e.g. shutting down the server when the extension is deactivated?

bryangingechen:
I was able to fix the issue I was having by adding a `.dispose()` method to the `Server` class. 

I also added a subscription to the `server.restarted` event in `InfoProvider`, but if the extension is activated by the opening of a `.lean` file, the server starts (and the event fires) before the `InfoProvider` is constructed, so nothing happens. I didn't immediately see a way to check if the server was already running -- any suggestions?

If I activate the extension by opening a markdown file, and then open a Lean file afterwards, then the info view will open automatically, as desired. By the way, I noticed that if I just open a markdown file, the status bar still says "Lean: checking visible files", even though the server isn't actually running.

bryangingechen:
OK, the most recent commit https://github.com/leanprover/vscode-lean/pull/105/commits/71a2c26fc4a891664c01d4ac74615b4e2ed8558d  fixed the issue which I was seeing. Previously, when `LeanSyncService.syncDoc` went to sync a Lean file and the server wasn't already running, it would start the server and then return before synching that file. I deleted the return and things are good now.

I also see how to check if the server is running, so the auto-open should work properly in the next commit.

bryangingechen:
Before the most recent commit, if you had a markdown file open and then clicked on a Lean file from the explorer, the auto-opened info view would show "No Lean file active"; presumably because the newly-opened editor didn't get marked as `activeTextEditor` until after the `autoOpen` function had already run. Adding a subscription to `onDidChangeActiveTextEditor` seemed like a reasonable way of fixing this.

gebner:
LGTM.

PR #109, opened 2019-01-16
Back and forward buttons for Lean documentation view

EdAyers:
A quick fix for the _Lean Documentation_ panel that presents a back and forward button that allow the user to navigate though the lean prover documentation.

### DIFF
--- a/src/docview.ts
+++ b/src/docview.ts
@@ -11,14 +11,15 @@ export function mkCommandUri(commandName: string, ...args: any): string {
 
 export class DocViewProvider implements Disposable {
     private subscriptions: Disposable[] = [];
-    private currentHtml : string = "";
-    private backstack : string[] = [];
-    private forwardstack : string[] = [];
+    private currentHtml: string = '';
+    private backstack: string[] = [];
+    private forwardstack: string[] = [];
     constructor() {
         this.subscriptions.push(
             commands.registerCommand('lean.openDocView', (url) => this.open(url)),
+            commands.registerCommand('lean.backDocView', () => this.back()),
+            commands.registerCommand('lean.forwardDocView', () => this.forward()),
             commands.registerCommand('lean.openTryIt', (code) => this.tryIt(code)),
-            //commands.registerCommand("workbench.action.navigateBack", () => this.back()),
         );
     }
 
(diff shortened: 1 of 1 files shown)


src/docview.ts
```diff
@@ -69,7 +73,27 @@ export class DocViewProvider implements Disposable {
                 link.attribs.href = mkCommandUri('lean.openDocView', href);
             }
         }
-        this.getWebview().webview.html = $.html();
+        $(`<nav style="box-sizing: border-box; position : sticky; top : 0px; padding : 4px; background : white">
```
gebner: Maybe we could have back/forward as buttons in the title bar instead?  You just need to add the commands to the `editor/title` menu in `package.json`: https://github.com/leanprover/vscode-lean/blob/master/package.json#L358

Then `"when"` clauses are a bit broken since we've switched to webview (from the previous previewHTML command).  For example here they should only show the button on infoview tabs: https://github.com/leanprover/vscode-lean/blob/master/package.json#L375  This does not work now (`resourceScheme` doesn't have the right value).

We'd need to set a context key like the markdown extension now, that indicates whether the docview is active or not: https://github.com/Microsoft/vscode/blob/1b61a70042a75e5f7273227b0974ffef51894adc/extensions/markdown-language-features/package.json#L104-L108 https://github.com/Microsoft/vscode/blob/master/extensions/markdown-language-features/src/features/previewManager.ts#L152-L154

Unfortunately, this option has the disadvantage that you need to click on the docview tab first, otherwise the buttons are not shown.

src/docview.ts
```diff
@@ -69,7 +73,27 @@ export class DocViewProvider implements Disposable {
...
+        <a class="btn" href="${mkCommandUri('lean.backDocView')}">back</a>
+        <a class="btn" href="${mkCommandUri('lean.forwardDocView')}">forward</a>
```
gebner: The `<nav>` bar has the advantage that it is always visible, which is nice.

 * Stylistically, how about `padding: 1ex; text-align: right;` for `<nav>`?
 * The buttons are not visually distinct from the content.  Maybe `border-bottom: 1px solid gray` or something?
 * Can we make the `<nav>` 100% in width?
 * Can we disable the back/forward links if the history is empty?
 * How about arrows in the links, i.e. `← back`, `forward →`?
 * The `<nav>` is shadowed by the `<pre>` blocks.

gebner: LGTM.  I'm not sure if the title bar buttons are worth the trouble since they are only visible when you focus the documentation view.

If you have time, could you look into adding a URL bar?  This was the other feature that [Patrick really missed](https://leanprover.zulipchat.com/#narrow/stream/179818-Lean-Together.202019/topic/Guinea.20pigs.20wanted!/near/154947296).

PR #110, opened 2019-01-17
Update translations.json

jcommelin:
This symbol was locked into core, but was recently liberated. Finally functors will have the symbol they deserve!


gebner:
See [discussion on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Vscode.20Translations.20file/near/155356990) and [other discussion](https://leanprover.zulipchat.com/#narrow/stream/116395-maths/topic/refactor.20functor.20notation/near/155361713).

jcommelin:
Ok, so given those discussions maybe this shouldn't be merged. Currently the emacs and VScode `\functor` translations are out of sync. This PR wanted to switch VScode over to the emacs convention. Maybe some emacs user should now attempt the converse direction (-;

fpvandoorn:
I think this can be closed?

Issue #95, opened 2018-10-16
"Install lean using elan" in extra places


khoek:
There is a bug where, when the extension generates any little "popup" box in the bottom corner of the screen (say the warning for the case when leanpkg.path has been updated and you are prompted as to whether you want to reload), "Install lean using elan" is presented as an option.

khoek:
(As opposed to just when lean can't be found by the extension.)

Another case is when lean crashes due to the delivery of a signal, and the extension prompts you to "Restart lean".

bryangingechen:
Thanks for the report; this has been fixed in 8fa6705.

PR #111, opened 2019-02-01
fix: restore functionality to offer to install elan

semorrison:


bryangingechen:
The idea behind setting `hasLean = true` there was to prevent the "install elan" dialogue from popping up unnecessarily (see #95). Ideally it should be at the first location in the code which is unreachable unless a working copy of lean is present (for example, I had thought that if there was no working lean, then by this point execution will have passed to the `catch` block). Where it is now seems to be too early since it prevents the dialog from ever showing up. It could be that this is the wrong approach entirely, but if not, can you check to see if putting that line later on in that block works better?

gebner:
Merging this now since I'm going on vacation tomorrow.  Feel free to submit another PR that fixes this properly.

PR #112, opened 2019-02-01
better fix for the elan prompt

semorrison:
This is a better fix for #95.


bryangingechen:
I just tested this and can confirm this fixes #95.

PR #114, opened 2019-02-13
always display number of goals

bryangingechen:
Re: this suggestion of @jcommelin https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/VScode.20extension/near/158436337


bryangingechen:
Changed "0 goals\n proof completed" to "goals accomplished" per @jcommelin's suggestion.

PR #116, opened 2019-03-21
fix(offer to install elan)

semorrison:
Sorry this has taken a few attempts. Hopefully this is better tested than the last one.

I've also cleaned up some of the messages, removing redundant information ("Lean: " prefixes in message boxes that say "Lean" anyway) and making the distinction between error messages and advice clearer.


semorrison:
@bryangingechen [confirmed](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/How.20to.20install.20Lean/near/161321160) that he's tested this PR on windows too.

PR #117, opened 2019-04-01
fix attempts to iterate over "undefined"

bryangingechen:
This PR fixes two error messages that I noticed which arise from attempts to iterate over undefined variables:

In `completion.ts`, sometimes `message.completions` is undefined; I've added a check against this.

In `holes.ts`, when the `LeanHoles` object is constructed, `this.holes` is undefined, leading to an error in `provideCodeActions`. I've initialized it to `[]`.


gebner:
Thank you so much!

PR #118, opened 2019-04-01
feature: display type in status bar item

bryangingechen:
This adds a status bar item (on by default, configurable with `lean.typeInStatusBar`) which displays the type information for the term underneath the cursor.

This addresses [a request](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/VScode.20extension/near/161738201) of @jcommelin. 

I first tried implementing this new status bar item in a new class, but I found that the info view provider's frequent server requests were interrupting those made by the new class. So, I've placed this feature in `infoview.ts` so that the status bar item can reuse the info from the requests that are already being made to the lean server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ We currently support a variety of features.
 * Fill in `{! !}` holes using <kbd>ctrl</kbd>+<kbd>.</kbd>
 * Tasks for leanpkg (<kbd>ctrl</kbd>+<kbd>p</kbd> task configure)
 * Tactic state filtering with regex
+* Type of the term under the cursor can be displayed in the status bar
 
 ## Requirements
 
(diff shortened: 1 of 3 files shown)


gebner:
> I first tried implementing this new status bar item in a new class, but I found that the info view provider's frequent server requests were interrupting those made by the new class.

That would have been my first question.  I completely forgot about this server limitation.

I find it a bit irritating to see `Type:` in the status bar even if there is no type at the moment.  How about dropping the `Type:` prefix in this case, or even entirely?

bryangingechen:
Done.

bryangingechen:
One more thing I should point out - if the live updating of the info view is paused, the new status bar item also won't be updated. It should be possible to make the behavior independent, but there would be some additional complexity in `updatePosition`.

jcommelin:
> One more thing I should point out - if the live updating of the info view is paused, the new status bar item also won't be updated. It should be possible to make the behavior independent, but there would be some additional complexity in `updatePosition`.

It might be good to document this in the code.

PR #119, opened 2019-04-14
remove hover decorations on rerender

bryangingechen:
This fixes [an issue](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/dreaded.20red.20blocks.20of.20death.20are.20back/near/163319685) reported by @kbuzzard on Zulip: the red blocks which appear when hovering over errors in the info view can get stuck if the code is edited before the mouse is removed.


gebner:
Thanks, this has been bothering me for years now as well!

PR #122, opened 2019-06-10
tactic state colorization fix

bryangingechen:
Fixes a coloring issue reported by @kckennylau [on zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/VScode.20extension/near/167727022):


digama0:
how about matching a single word followed by a colon rather than blacklisting certain characters? You missed the {{ bracket as well.

bryangingechen:
A single word won't work since the tactic state sometimes combines multiple variables together if they have the same type, e.g.
Screenshot 2019-06-09 23 32 02

Also I think that blacklisting is used here since in Javascript regexes, "word" characters don't include unicode symbols. I'll add the other brackets shortly.

Issue #120, opened 2019-05-31
tab-completion should show all types, and no question marks


semorrison:
Tab-completion in VS-Code only [shows the type of the selected option, and prepends a `?` to each variable](https://leanprover.zulipchat.com/user_uploads/3121/ykVEWSMsfHNpO985b-l2W2Lm/Screenshot-2019-05-31-16.03.04.png). In emacs it's much nicer: [the type is displayed for every option, and instead of `?` the variables are highlighted](https://leanprover.zulipchat.com/user_uploads/3121/cNzLaymd1BAPcVCzoolsRQaN/Screenshot-from-2019-05-31-16-22-23.png).

gebner:
When I click on the links I only see `You are not authorized to view this file.`  Can you upload the screenshots (?) somewhere else?  (e.g. on github, drag&drop into the issue should work.)

semorrison:
<img width="604" alt="Screenshot-2019-05-31-16 03 04" src="https://user-images.githubusercontent.com/477956/58752344-32e74b80-84a5-11e9-8d81-701e8d1a1535.png">

![Screenshot-from-2019-05-31-16-22-23](https://user-images.githubusercontent.com/477956/58752346-38dd2c80-84a5-11e9-8896-40bed9f976b7.png)

bryangingechen:
It's possible to remove the question marks and to display the types (and tactic params) in the CompletionItem's `label` property, but VS code doesn't support any kind of formatting. [Here's a branch](https://github.com/bryangingechen/vscode-lean/tree/better-completions) where I tried to adjust the completions towards what [the emacs mode does](https://github.com/leanprover/lean-mode/blob/master/company-lean.el)). I'm not sure if I like the result better since it's not as clear to me what will end up inserted.

semorrison:
@bryangingechen, how about we use your branch, but drop the `.replace(/\?([^\s]+)/g, '$1');`?

That way the resulting type will be shown, and I agree if we can't have highlighting then the question marks should just be shown.

bryangingechen:
To be honest, I actually prefer it the other way around, with no question marks and no types displayed by default. However, I'll update my branch as you describe and put the types behind a config option.

semorrison:
Sorry, I didn't see this response before making my other post. I'll admit
to being mystified why you'd prefer not to see the types, but gusta non
disputandum!

bryangingechen:
[...]

PR #123, opened 2019-06-25
Better completions

bryangingechen:
Adds new config option `lean.typesInCompletionList`, which controls whether the types of all items in the list of completions are displayed. Currently, only the type of the highlighted item is shown and this will continue to be the default.

Addresses #120, as suggested by @semorrison.


semorrison:
Could we change the default to show all types? I'm actually not
understanding why you'd ever want the other behavior.

bryangingechen:
I personally thought that the completion list was too cluttered with all the types on by default, and I also was concerned that having extra text that doesn't get inserted when the completion item is selected could be confusing. Thus, I decided to keep the default behavior the same to be safe. However, I don't really mind if it's changed.

PR #127, opened 2019-08-09
add input translation from \f<> to ‹›

ratmice:
I haven't tried building this, but edited the translations.json locally,
ran with vscode and then applied the diff,
this has only been lightly tested within vscode itself as well, but seemed to work.

Anyhow it would be nice if we could save a few keystrokes inputting \f< \f>, by combining them into \f<>.

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -77,6 +77,7 @@ class TextEditorAbbrevHandler {
             [this.leader + '[[]]']: '⟦⟧',
             [this.leader + '<>']: '⟨⟩',
             [this.leader + '([])'] : '⟮⟯',
+            [this.leader + 'f<>']: '‹›',
         };
         if (range) {
             const replacement = hackyReplacements[this.editor.document.getText(range)];


bryangingechen:
I think it should be added here instead:

https://github.com/leanprover/vscode-lean/blob/d3d05c066b49c1f23f6524e7b3a5ce293bdb926f/src/input.ts#L75-L80

That way, the cursor will automatically be placed between the brackets when you hit tab.

ratmice:
> I think it should be added here instead:
> 
> https://github.com/leanprover/vscode-lean/blob/d3d05c066b49c1f23f6524e7b3a5ce293bdb926f/src/input.ts#L75-L80
> 
> That way, the cursor will automatically be placed between the brackets when you hit tab.

Thanks, I modified this to as per your comment,

I had a bit of difficulty figuring out how to test it, trying to run via F5, seems to throw a bunch of errors (even without the modification), but I finally managed to test it running code --extensionDevelopmentPath=`pwd`

The changes you requested seemed to work better, and put the cursor in between the brackets.

jcommelin:
Hooray! :tada:

PR #135, opened 2019-12-31
exclude angle brackets in wordPattern

kckennylau:
per https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/VScode.20extension/near/184537385


gebner:
Did you test this?

kckennylau:
I haven't.

bryangingechen:
I just did a little testing and it seemed to work for me. Only suggestion I'd make is to add the following characters as well: `⦃⦄⟦⟧⟮⟯‹›`.

kckennylau:
done

PR #143, opened 2020-02-14
tactic suggestions

benti:
Copy and paste output from tactics like `squeeze_simp`.

### DIFF
--- a/src/tacticsuggestions.ts
+++ b/src/tacticsuggestions.ts
@@ -1,4 +1,4 @@
-import { commands, Disposable,  window } from 'vscode';
+import { commands, Disposable,  window, TextEditor, TextEditorEdit } from 'vscode';
 import { Server } from './server';
 
 /** Pastes suggestions provided by tactics such as `squeeze_simp` */
@@ -7,28 +7,22 @@ export class TacticSuggestions implements Disposable {
     constructor(private server: Server) {
         const command = 'lean.pasteTacticSuggestion';
 
-        const commandHandler = (name: string = 'world') => {
-            const editor = window.activeTextEditor;
+        const commandHandler = (textEditor : TextEditor, edit : TextEditorEdit) => {
+            const curFileName = textEditor.document.fileName;
+            const curPosition = textEditor.selection.active;
+            const curSelection = textEditor.selection;
+            const messages = this.server.messages
+                .filter((m) => m.file_name === curFileName &&
+                    m.pos_line === curPosition.line + 1)
+                .sort((a, b) => a.pos_col - b.pos_col);
 
-            if (editor) {
-                const curFileName = editor.document.fileName;
-                const curPosition = editor.selection.active;
-                const curSelection = editor.selection;
-                const messages = this.server.messages
-                    .filter((m) => m.file_name === curFileName &&
-                        m.pos_line === curPosition.line + 1)
-                    .sort((a, b) => a.pos_col - b.pos_col);
-                // console.log(messages)
-                if (messages.length > 0) {
-                    editor.edit(editBuilder => {
-                        editBuilder.replace(curSelection, messages[0].text);
-                    });
-                }
+            if (messages.length > 0) {
+                edit.replace(curSelection, messages[0].text);
             }
         };
 
         this.subscriptions.push(
-            commands.registerCommand(command, commandHandler)
+            commands.registerTextEditorCommand(command, commandHandler)
         );
     }
 
--- a/src/tacticsuggestions.ts
+++ b/src/tacticsuggestions.ts
@@ -1,5 +1,6 @@
-import { commands, Disposable,  window, TextEditor, TextEditorEdit } from 'vscode';
+import { commands, Disposable,  window, TextEditor, TextEditorEdit, Position, Range } from 'vscode';
 import { Server } from './server';
+import { Message } from 'lean-client-js-node';
 
 /** Pastes suggestions provided by tactics such as `squeeze_simp` */
 export class TacticSuggestions implements Disposable {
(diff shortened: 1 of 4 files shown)


benti:
https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/VScode.20message.20window

src/tacticsuggestions.ts
```diff
@@ -0,0 +1,38 @@
+import { commands, Disposable,  window } from 'vscode';
+import { Server } from './server';
+
+/** Pastes suggestions provided by tactics such as `squeeze_simp` */
+export class TacticSuggestions implements Disposable {
+    private subscriptions: Disposable[] = [];
+    constructor(private server: Server) {
+        const command = 'lean.pasteTacticSuggestion';
+
+        const commandHandler = (name: string = 'world') => {
```
gebner:
```suggestion
        const commandHandler = (name: string) => {
```

src/tacticsuggestions.ts
```diff
@@ -0,0 +1,38 @@
...
+
+            if (editor) {
+                const curFileName = editor.document.fileName;
+                const curPosition = editor.selection.active;
+                const curSelection = editor.selection;
+                const messages = this.server.messages
+                    .filter((m) => m.file_name === curFileName &&
+                        m.pos_line === curPosition.line + 1)
+                    .sort((a, b) => a.pos_col - b.pos_col);
+                // console.log(messages)
```
gebner:
```suggestion
```

src/tacticsuggestions.ts
```diff
@@ -0,0 +1,38 @@
...
+                if (messages.length > 0) {
+                    editor.edit(editBuilder => {
+                        editBuilder.replace(curSelection, messages[0].text);
+                    });
+                }
+            }
+        };
+
+        this.subscriptions.push(
+            commands.registerCommand(command, commandHandler)
```
gebner:
```suggestion
            commands.registerTextEditorCommand(command, commandHandler)
```

This also passes the correct editor to the commandHandler function.

src/tacticsuggestions.ts
```diff
@@ -0,0 +1,38 @@
...
+                const curPosition = editor.selection.active;
+                const curSelection = editor.selection;
+                const messages = this.server.messages
+                    .filter((m) => m.file_name === curFileName &&
+                        m.pos_line === curPosition.line + 1)
+                    .sort((a, b) => a.pos_col - b.pos_col);
+                // console.log(messages)
+                if (messages.length > 0) {
+                    editor.edit(editBuilder => {
+                        editBuilder.replace(curSelection, messages[0].text);
```
gebner:
```suggestion
                        editBuilder.replace(???, messages[0].text);
```

The range to be replaced should be calculated from the message.  Maybe `message.pos_col` until the next comma or the end of the line, whichever comes first?

bryangingechen: Just looking for the first comma won't work for `squeeze_simp [lemma1, lemma2]` or `tidy? { tactics := [tactic1, tactic2] }`. The "end of the line" idea could also fail for a long list of simp lemmas broken up with linebreaks.

jcommelin: Maybe the `try this:` magic string should also include who sent the message. Something like `tidy says, try this:` and `simp says, try this`, etc... so that the replacement script has at least a little clue to look for what it should replace.

jcommelin: Oooh, don't Lean tactics know which position in the file they are in? So `squeeze_simp` and `suggest` and friends could output coordinates together with the `try this:` message

benti: I implemented till the end of line for now.

bryangingechen:
It might be nice to register this as ["CodeActionProvider"](https://code.visualstudio.com/api/references/vscode-api#CodeActionProvider) in addition to the custom command. Then VS code would display a light bulb next to the tactic as a visual indication that there's an action to take. Clicking that light bulb (or hitting <kbd>ctrl</kbd>+<kbd>.</kbd>) would open the "code actions" menu. This could be a way of choosing a suggestion from tactics that return more than one (like `suggest`). 

For an example of this API in use, see [holes.ts](https://github.com/leanprover/vscode-lean/blob/master/src/holes.ts#L81).

benti:
> It might be nice to register this as ["CodeActionProvider"](https://code.visualstudio.com/api/references/vscode-api#CodeActionProvider) in addition to the custom command. Then VS code would display a light bulb next to the tactic as a visual indication that there's an action to take. Clicking that light bulb (or hitting ctrl+.) would open the "code actions" menu. This could be a way of choosing a suggestion from tactics that return more than one (like `suggest`).

I think it would be even nicer to click on the "Try this" or on the suggestions directly to insert them, as Gabriel proposed. Of course we could have the light bulbs in addition.

bryangingechen:
> I think it would be even nicer to click on the "Try this" or on the suggestions directly to insert them, as Gabriel proposed.

Yes, agreed. One way to achieve this is to add a transformation step [in `renderMessages` in infoview.ts](https://github.com/leanprover/vscode-lean/blob/master/src/infoview.ts#L491) which acts on messages with a "severity" of "information" and wraps any "Try this" with an `<a>` tag. The `href` attribute could then contain a `command:` link, similar to the one defined on line 488, except calling something like 'lean.pasteTacticSuggestion'.

benti:
The suggestions can now be clicked to insert them!

bryangingechen:
I haven't had a chance to test this out yet, but assuming it works, I'd be happy merging this as is, as soon as the associated mathlib PR lands. I can try adding the CodeActionProvider stuff in a separate PR if it's too much trouble to do here.

src/tacticsuggestions.ts
```diff
@@ -0,0 +1,95 @@
...
+            new Position(line, col),
+            new Position(line,
+                textEditor.document.lineAt(line).range.end.character)
+        )
+        await textEditor.edit(editBuilder => {
+            editBuilder.replace(range, suggestion)
+        });
+
+        // Focus text editor
+        const lastCol = textEditor.document.lineAt(line).range.end.character;
```
gebner: This part is only relevant for the info view, right?  In that case, it should only be executed by the info view command.

benti: It also jumps to the last character on the line after pressing `alt+v`. But yes, I could implement this differently for each command.

gebner: Ah ok, then it's fine.

PR #145, opened 2020-02-29
infoview: delete "Updating" label, make title reflect DisplayMode

bryangingechen:
See Zulip https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/noob.20question%28s%29/near/189396573

I also added a title attribute to the tactic state filter.

### DIFF
--- a/media/infoview.css
+++ b/media/infoview.css
@@ -43,7 +43,7 @@ div#filter {
   display: block;
   position: absolute;
   top : 0px;
-  right: 90px;
+  right: 30px;
   width: auto;
   padding: 3px 0px 1px 0px;
   z-index: 2;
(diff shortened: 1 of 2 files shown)


bryangingechen:
@gebner Any comments on this one?

media/infoview.css
```diff
@@ -43,7 +43,7 @@ div#filter {
   display: block;
   position: absolute;
   top : 0px;
-  right: 90px;
+  right: 30px;
```
gebner: There's probably a better way to do this in CSS, but that's for another PR.

gebner:
@bryangingechen Thanks for pinging me!

PR #146, opened 2020-03-01
add code actions for tactic suggestions

bryangingechen:
I use the diagnostics provided by the Lean server to add code actions (the "light bulb" menu) for the "Try this" tactic suggestions.

I also made some slight style edits and fixed a minor bug in `pasteIntoEditor`: previously if the user hit alt+v away from a tactic suggestion, `m.text.match(new RegExp(this.regex, 'm'))` would be `null`, so `m.text.match(new RegExp(this.regex, 'm'))[1]` would throw an error.

Finally, I updated the README and did some other cleanup there, e.g. formatting tweaks + adding links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,18 +12,21 @@ We currently support a variety of features.
 * Auto-completion based on context and type via the Lean server
 * Error messages / diagnostics
 * Customizable Unicode input support (e.g. type `\la`+<kbd>tab</kbd> to input `λ`)
-* Info view window to show goal, tactic state, and error messages (<kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>enter</kbd> for local goal view or <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>alt</kbd>+<kbd>enter</kbd> for all messages)
+* Info view window to show goal, tactic state, and error messages:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -12,18 +12,21 @@ We currently support a variety of features.
 * Auto-completion based on context and type via the Lean server
 * Error messages / diagnostics
 * Customizable Unicode input support (e.g. type `\la`+<kbd>tab</kbd> to input `λ`)
-* Info view window to show goal, tactic state, and error messages (<kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>enter</kbd> for local goal view or <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>alt</kbd>+<kbd>enter</kbd> for all messages)
+* Info view window to show goal, tactic state, and error messages:
+  - click <img src="media/display-goal-light.svg"> or hit <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>enter</kbd> for local goal view, or
```
gebner: BTW, you can't use SVGs in the readme of a vscode extension.  I didn't know this either.  I've now changed the SVGs to PNGs: https://github.com/leanprover/vscode-lean/commit/e3b5a822291a2ebc5d4358dab41e4a602c211c90

bryangingechen: Ah, OK. VS Code gave me a warning about those, but I wasn't sure what it was referring to.

PR #147, opened 2020-03-06
fix range in diagnostics:

bryangingechen:
Previously, if the position of a message from the server
happened to lie between "words" (e.g. at the end of a file),
`getWordRangeAtPosition` would return undefined, leading
to a broken range. This change adds a fallback to the old
behavior of creating a `Range` that both begins and ends
at the position of the server message.


bryangingechen:
(For future reference, there are a few diagrams about `getWordRangeAtPosition` here: https://github.com/Microsoft/vscode/issues/4694#issuecomment-283380455)

Issue #151, opened 2020-03-17
Add button to info view to track current position


gebner:
Right now the info view always shows the error messages at the cursor position, but sometimes you want to monitor the error messages from somewhere else.  It would be nice if you could tell the info view to track/focus/bikeshed one position and then continue to work somewhere else.  You can already pause the info view, but then you unfortunately don't get any updates.

As requested by @cipher1024:
> One example of that use is that you write a theorem and you try to tweak the tactics to make progress on it. You place your focus on that theorem, then you start changing the code of the tactic. You have one panel that gives you the error in that tactic and another that shows you what is going on at the focus. You don't have to keep going back and forth. It's a bit like having an `#eval` right after the code that you're writing

cipher1024:
It would help to do that to have the flag that you mentioned:

> Probably the easiest way to support this concrete use case is to add a flag to the Lean 3 server so that it always loads the olean files, even if they're out of date.

PR #158, opened 2020-05-08
Add "sticky position" option

digama0:
Basic implementation of #151. The UX is pretty terrible right now, but the functionality is there. Use the `Info View: Toggle Sticky Position` command to mark the current location of the cursor as the destination of future goal requests. Afterwards the cursor is free to move without affecting the goal, and the goal will continue to be updated. The mark also rides along with text edits as you would expect. Toggling the option again returns to normal behavior.

At minimum, I think there should be a button indicating the status of this setting and allowing you to change it, say next to the pause button. But I'm no designer, I will leave that to others.


gebner:
This is wonderful!

gebner:
I've merged this for now even though I have no time to improve the UI (add a button).  Anybody is free to submit a follow-up PR.

In the medium run, I hope @EdAyers will include this feature in #159.

digama0:
The position tracking here is also pretty broken, and I haven't managed to find a very good replacement. (To test, try setting the mark and then typing text on the line above it or just before it, and see if it holds its position.)

EdAyers:
Hi gang, I've implemented web-view-side pinning on my widget branch. It doesn't look pretty but the functionality is there. I changed the web-view so that it shares a Connection with the server. So now you can pin infoviews. No position tracking yet

PR #159, opened 2020-05-10
Infoview overhaul

EdAyers:
I rewrote the infoview code so that it is a full typescript/react project. It is also 'widget ready'.
There are still some tasks to be done:

- [x] ~filtering hypothesis lists~ --> should be implemented in lean
- [x] 'pinning' infoviews at certain positions

Anyway the main purpose of this PR is to just be a place where we can discuss what needs to be done here before a merge.


package.json
```diff
@@ -427,9 +427,12 @@
 	],
 	"main": "./out/src/extension",
 	"scripts": {
-		"vscode:prepublish": "tsc -p ./",
-		"compile": "tsc -watch -p ./",
-		"lint": "eslint -c .eslintrc.js src/*.ts"
+		"vscode:prepublish": "tsc -p ./ && webpack --env.production",
+		"compile-infoview": "webpack --env.development",
+		"compile": "tsc -watch -p ./ && webpack --env.development --watch",
```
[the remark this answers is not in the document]
EdAyers: I've migrated everything to webpack now so it can build / watch both simultaneously.

gebner:
I took another look since you advertised it on the other thread.  This is starting to look great!

 - I'm not sure what I think of the collapsible messages.  I imagine that some users will collapse them on accident.  But maybe they're useful.
 - It's awesome that the infoview shows *both* the pinned and the current location.
 - Newlines in the error messages are ignored.
 - The info view should only show errors on the current line.  There is also the "info view all errors on line" option that I have always enabled, but I guess some people want to have the default.
 - I don't see any goal at all.

EdAyers:
>     * I don't see any goal at all.

What build of lean was this on?

gebner:
```
$ lean --version
Lean (version 3.13.2, commit f36f98fba8e7, Release)
```

I don't think the relevant part of the server protocol has changed in years, though.

EdAyers:
@gebner 
>     * I'm not sure what I think of the collapsible messages.  I imagine that some users will collapse them on accident.  But maybe they're useful.

Yeah there have definitely been times in the past where I wanted collapsible messages. They should be straightforward to remove so maybe we should do a poll.

>     * It's awesome that the infoview shows _both_ the pinned and the current location.

Thanks! It now also corrects the pinned location if you insert text. It's using @digama0 's algorithm.

>     * Newlines in the error messages are ignored.

This is fixed now I was using `white-space: normal` instead of `white-space: pre-normal`.

>     * The info view should only show errors on the current line.  There is also the "info view all errors on line" option that I have always enabled, but I guess some people want to have the default.

Fixed.

EdAyers:
I think the issue of not showing goals is fixed now. It was conditioning showing the goals on the DisplayMode

gebner:
I'm seeing the goal now, but it keeps showing the goal even if I click on another line (which doesn't have a goal).

gebner:
The expanded state of the messages widget is not persisted.  Say I have one error in a file and collapse both the messages and all messages widgets.  If I then remove the error and add the error back, then both widgets are expanded again.

I'd also prefer to have "all messages" collapsed by default.  I believe that's also the current default.

media/infoview.css
```diff
@@ -12,19 +12,19 @@ body {
   padding: 0px;
 }
 
-div
+/* div
```
gebner, line 15: Please remove commented-out code.

src/infoview.ts
```diff
@@ -136,37 +113,50 @@ export class InfoProvider implements Disposable {
...
-                    }
-                    this.updatePosition(false);
-                } else {
-                    this.stickyPosition = true;
-                    const pos = editor.selection.active;
-                    editor.setDecorations(this.stickyDecorationType, [new Range(pos, pos)]);
-                    this.updatePosition(false, pos);
-                }
-            }),
+            // commands.registerTextEditorCommand('lean.infoView.copyToComment',
```
gebner, line 135: These commands should be made to work again.

infoview/messages.tsx
```diff
@@ -0,0 +1,85 @@
...
+        // JSON.stringify([Uri.file(m.file_name), m.pos_line, m.pos_col]));
+        // JSON.stringify([(m.file_name), m.pos_line, m.pos_col])); // [TODO] Uri.file isn't available in the webview?
+    const shouldColorize = m.severity === 'error';
+    let text = escapeHtml(m.text)
+    text = shouldColorize ? colorizeMessage(text) : text;
+    const title = `${b}:${l}:${c}`;
+    return <details open>
+        <summary className={m.severity + ' mv2 pointer'}>{title}</summary>
+        <div className="ml3">
+            <pre className="font-code" style={{whiteSpace: 'pre-wrap'}} dangerouslySetInnerHTML={{ __html: text }} />
```
gebner, line 30: This doesn't enable line-wrapping for me, I still need to scroll horizontally:
```lean
example (h : 0 = by iterate 100 { refine _ * 1 }; exact 0) : true :=
by trivial
```

Somewhat more disturbingly, if I replace `trivial` by `trivia`, then I see *old* error messages from the line above ("unknown identifier 'exac'", etc.).

src/server.ts
```diff
@@ -132,7 +127,7 @@ export class Server extends leanclient.Server {
                     // json parsing errors
                     if (e.message.startsWith('cannot parse: ')) {
                         stderrOutput.append(e.message + '\n');
-                        stderrOutput.show();
+                        // stderrOutput.show();
```
gebner, line 130: :+1:

infoview/messages.tsx
EdAyers: Yes I'm also getting a bug where it is accumulating old error messages

EdAyers: This is because I didn't set pre-wrap on the goal viewer in goal.tsx

gebner:
- There are merge conflicts.
- Can you reduce/remove the indentation?  I don't have that much screen real estate on my laptop.
- I'm getting lots of "error updating: interrupted".  We probably need to rework/remove the cancellation in the server.
error_interrupted
- Both the widget and the normal tactic state are shown simultaneously.

EdAyers:
Does anyone know why this is failing: https://github.com/EdAyers/vscode-lean/runs/751465242?check_suite_focus=true ?

gebner:
@EdAyers Are there any obvious bugs left?  Otherwise I think it's time to merge this PR.

PatrickMassot:
What about releasing a beta version? You could post a link to the vsix on Zulip and let people play with it. The version that I got last Thursday was amazing but pretty far from being bug free, in a way that really make it hard to use.

bryangingechen:
Ed asked me to give my thoughts a few weeks ago and I'm just writing down a few examples of bugs / confusing behavior below.

---

Screenshot 2020-06-10 04 47 14

Why is there a "Widget" and a "Tactic state" which duplicate each other? Also, the spacing between lines seems to be too large in the widget and too small in the tactic state.

---

Screenshot 2020-06-10 04 50 00

If I switch to "Lean Messages" mode, there's duplication of error messages if my cursor is on the same line as the error.

---

The "pin" icon in the info view and the top right of the editor panel should be made consistent. Actually, I guess the pin icon in the editor panel should be removed, since now multiple things can be pinned / unpinned independently.

---

I think the spacing between icons in the info view can be reduced.

---

If I restart while the info view is in "Lean Messages" mode, the info view title starts out as "Lean Goal" but it still has "All messages", and gets stuck at "Waiting for info..." until I move my cursor. It seems to take two cursor movements before the info view populates.

---

It's not clear now what the "Copying the contents" button will send to the editor. In the old info view, it copies everything in the info view, but in the new info view it seems to only copy the contents of a tactic state widget. It'd be nice if we could copy error messages again too.

---

What's the status of the tactic state filtering options? Do they still work (just with no UI)?

---

The README needs an update. I'd be happy to contribute this.

gebner:
@PatrickMassot @bryangingechen Thanks for the comments and putting in the time to test this PR!

Some of the user interface issues are on the Lean side, where the tactic state widget is implemented.  The issues Bryan has raised about the spacing, etc., are now fixed in 3.16.

So we're going to 1) wait for mathlib to move to 3.16, and then 2) post a vsix on zulip to test.

EdAyers:
Yeah I think Patrick is right, we should release a beta in sync with 3.16. I don't think that lean 3.16 needs to be a beta though because it should work with all versions of the extension. I also think that we should pass 'widgets: true' in the info request so that we have the option to change the defaults in lean to not create widgets for each info request later.

EdAyers:
Also massive thanks to Bryan and Patrick

EdAyers:
re @bryangingechen 's tactic state filtering options query, this should be reimplemented in Lean as part of the tactic state widget.

bryangingechen:
Will the new info view still work well with older versions of Lean? For instance, I know there's a fix for the "interrupted" issues, in the newer versions of community Lean, but will users of older versions of Lean just have to deal with that by moving the cursor around manually?

EdAyers:
It should be backwards compatible. The 'cursor moving around' issues were just because it wouldn't bother to update if the infoview if the lean task status changed (which is fixed now).  The only snags I've found are that if you have multiple pinned infos they will not always update properly and you get a "Error updating: interrupted. Try again." message.

EdAyers:
Hmm ok there are a few more error updated cases than I thought

EdAyers:
I don't quite get what is causing the interruption errors, it seems random to me, but at least now it is not quite so kaleidoscopic when you drag the cursor around.

bryangingechen:
OK, thanks for looking into it. Hopefully there'll be some solution since it'd be a shame to degrade the experience of looking at projects relying on older versions of Lean.

EdAyers:
Yeah you are right it has to still be good for older projects

EdAyers:
The issues with interupted errors should be behind us now.

PR #161, opened 2020-05-15
cleanup commands, update README

bryangingechen:
Daniel Fabian pointed out [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/vscode.20extension/near/197605170) that the "toggle updates" command did not appear in the command palette. On investigation, it turned out that this was because the `when` condition was referring to an old `resourceSchema` condition which never holds. The same held true for the `lean.infoView.displayGoal` and `lean.infoView.displayList` commands. Furthermore, all of these commands were registered as textEditorCommands, so it was impossible to call them when the info view panel was focused. I fixed this condition following [a post on stackoverflow](https://stackoverflow.com/a/54917749).

I also figured that it might be helpful to have clickable icons for the "copy to comment" and "toggle sticky position" commands, so I "drew" some by fiddling with SVG path commands.

Finally, I removed an unused argument in the `updatePosition` function.


gebner:
Thanks!  The icons look great!  At first I didn't find the buttons (I looked next to the pause button), but I'm not sure I hate the location either.

bryangingechen:
Yeah, it was definitely much easier just to add them to the "editor/title" menu than to make new elements in the infoview. 

Now that I think of it, we should probably emphasize in the README or the command names themselves the distinction between the textEditorCommands (which can only be run when a text editor is focused) and the other commands.

PR #164, opened 2020-05-24
Update task config & ignore non-Lean regions of interest

Vtec234:
Does two small things:
- Brings the `tasks.json` configuration in line with the defaults for VSCode versions post 1.13. I had to do this because of the note [here](https://code.visualstudio.com/docs/editor/tasks#_convert-from-010-to-200) - since version 0.1.0 configurations are ignored in multi-root workspaces, VSCode will literally refuse to debug this extension if I also have a folder with, say, the C++ server open.
- Stops the extension from tracking regions-of-interest on files that aren't written in Lean. Currently, the Lean server stores ROIs for things like `leanpkg.toml`, `README.md`, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -467,8 +467,9 @@
 	],
 	"main": "./out/src/extension",
 	"scripts": {
-		"vscode:prepublish": "tsc -p ./",
-		"compile": "tsc -watch -p ./",
+		"vscode:prepublish": "npm run compile",
+		"compile": "tsc -p ./",
+		"watch": "tsc -watch -p ./",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -477,6 +478,7 @@
 		"cheerio": "^1.0.0-rc.3",
 		"express": "^4.17.1",
 		"hasbin": "^1.2.3",
+		"lean-client-js-core": "^1.2.12",
```
Vtec234, line 481: NPM doesn't seem to make this available just because it's a dependency of "lean-core-js-node", so I added it explicitly here.

PR #167, opened 2020-06-09
Add github action for linting and packaging.

EdAyers:
Seems sensible. We could also add a release step.


gebner:
I don't think there's a big need to optimize the release process, this works pretty smoothly at the moment using `./release.sh`.

PR #170, opened 2020-06-16
Add missing Unicode subscripts and superscripts

utensil:
I noticed some superscript capital, superscript minuscule and underscript minuscule letters are missing from `translations.json` and I need some of them when I define some notations, so I carefully filled all the gaps I know of and submitted this PR.

Reference: https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts


translations.json
```diff
@@ -1217,9 +1237,9 @@
     "^TM": "™",
     "_a": "ₐ",
     "_e": "ₑ",
+    "_h": "ₕ",
     "_i": "ᵢ",
     "_j": "ⱼ",
-    "_h": "ₕ",
```
utensil, line 1222: This is the only modification that's not only addition. I reorder the letters a bit so it follows the order of alphabet to for easy lookup.

gebner:
Let's hope that https://github.com/stevengj/subsuper-proposal/ gets proposed and accepted at some point in the future!

PR #175, opened 2020-06-17
Fix 173

### LINKED ISSUE #173: Error messages do not update in info view
1. Open the following file in vscode:
```
example : false := trivial
```
2. Close the info view.
3. Open the info view.
4. The error message is not shown.

cc @EdAyers

EdAyers:
I think fixes #173


gebner:
I think this fixes it for me.

PR #176, opened 2020-06-17
Remove DisplayMode

EdAyers:
The only thing that DisplayMode was doing was changing the title of infoview and deciding whether or not to display the 'all goals' collapsible section of the infoview. So I think it makes sense to remove it and to also remove the distinction between "Display Goals" and "Display Messages". 

Also remove config to do with filtering hypotheses.

### DIFF
--- a/infoview/index.tsx
+++ b/infoview/index.tsx
@@ -1,7 +1,7 @@
 import { global_server, post, PositionEvent, ConfigEvent, SyncPinEvent, PauseEvent, ContinueEvent, ToggleUpdatingEvent, TogglePinEvent } from './server';
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
-import { ServerStatus, Config, defaultConfig,  Location, locationKey, locationEq, DisplayMode } from '../src/shared';
+import { ServerStatus, Config, defaultConfig,  Location, locationKey, locationEq } from '../src/shared';
 import { Message } from 'lean-client-js-core';
 import './tachyons.css' // stylesheet assumed by Lean widgets. See https://tachyons.io/ for documentation
 import './index.css'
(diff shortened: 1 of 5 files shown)


EdAyers:
I'll ask on zulip whether this is sensible

package.json
```diff
@@ -98,65 +98,11 @@
...
-					"type": "boolean",
-					"default": true,
-					"description": "Info view: auto open shows goal and messages for the current line (instead of all messages for the whole file)"
-				},
 				"lean.infoViewStyle": {
 					"type": "string",
 					"default": "",
 					"description": "Add an additional CSS snippet to the info view."
 				},
-				"lean.infoViewTacticStateFilters": {
```
gebner, line 111: These still work.

package.json
```diff
@@ -366,18 +276,6 @@
 				"key": "alt+v",
 				"mac": "alt+v",
 				"when": "editorTextFocus && editorLangId == lean"
-			},
-			{
-				"command": "lean.infoView.displayGoal",
-				"key": "ctrl+shift+enter",
```
gebner, line 372: This command (and in particular the keycombo) should continue to work.

package.json
```diff
@@ -98,65 +98,11 @@
 					"default": true,
 					"description": "Info view: open info view when Lean extension is activated."
 				},
-				"lean.infoViewAutoOpenShowGoal": {
```
gebner, line 101: This should remain configurable.

package.json
EdAyers: Yes it does still work using `lean.displayGoal` instead of `lean.infoview.diaplayGoal`

infoview/index.tsx
```diff
@@ -94,7 +94,7 @@ function Main(props: {}) {
                 const isCursor = locationEq(loc,curLoc.loc);
                 return <Info loc={loc} isPaused={paused} setPaused={setPause(i)} key={i} isPinned={true} isCursor={isCursor} onEdit={onEdit} onPin={unpin(i)}/>}) }
             {!isPinned(curLoc.loc) && <Info loc={curLoc.loc} isPaused={curLoc.paused} setPaused={setPause()} key={pinnedLocs.length} isPinned={false} isCursor={true} onEdit={onEdit} onPin={pin}/>}
-            <details>
+            <details open={!config.infoViewAutoOpenShowGoal}>
```
gebner, line 97: Is the `<Messages/>` here rendered always, even if collapsed?  Is this a performance issue?

EdAyers: Yes it is rendered always. Afaik it's not a perf issue.

PR #179, opened 2020-06-17
Replace horizontal bar with title colour

EdAyers:
Also remove 'Widget' container because it takes up space.


gebner:
Feels much more peaceful now.

PR #182, opened 2020-06-19
Add support for select boxes.

EdAyers:
Allows 'onChange' to be called on `<select>` html elements. Which means that we can support dropdown boxes (helper function in lean tbc).

Eg:

image

### DIFF
--- a/infoview/goal.tsx
+++ b/infoview/goal.tsx
@@ -28,6 +28,7 @@ export function Goal(props: GoalProps): JSX.Element {
         {reFilters.length !== 0 && <div className="fr">
             <label>filter: </label>
             <select value={filterIndex} onChange={(e) => setFilterIndex(e.target.value)}>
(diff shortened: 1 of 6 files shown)


bryangingechen:
Older versions of Lean still won't get a tactic state filter, right?

EdAyers:
Yep see https://github.com/leanprover-community/lean/pull/346 for the filtering

EdAyers:
It should be not too hard to re-add the filtering menu for the non-widget tactic state, I think that all of the filtering code is still there its just the dropdown is gone.

EdAyers:
Could you post some examples of `infoViewTacticStateFilters` config entry so I can test it?

bryangingechen:
Hmm, I think I've only used the two filters in `package.json`. Maybe we could ask in Zulip?

PR #184, opened 2020-06-23
Support widget trace messages.

gebner:
There are still some performance problems.  In particular, all widgets in the messages are rendered at all times...

@EdAyers Do you know if there is a standard way to make `<details>` only render (on the react side) the visible elements?


EdAyers:
I'll figure it out

EdAyers:
See #185

PR #185, opened 2020-06-23
Widgets only fetch widget info when they are visible in the viewport.

EdAyers:
Fix performance issue where many messages with widgets would impact performance. Now it will only ask the server for widget info when it becomes visible in the viewport. So this also means that if there is a very long list of messages, opening the All Messages view should remain performant because it will only load widgets for messages that are scrolled in to view.

### DIFF
--- a/infoview/widget.tsx
+++ b/infoview/widget.tsx
@@ -60,9 +60,34 @@ class WidgetErrorBoundary extends React.Component<{children},{error}> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


gebner:
Ah, this is an interesting solution, thanks!

It gives even more information than I would have expected.  The `isVisible` property even reflects whether the element is in the current viewport, and not just if it is `display: none`.

AFAICT one issue (just so that I don't forget) is that when a widget scrolls out of the viewport and back in, then it will re-send the `get_widget` command.

PR #204, opened 2020-07-08
Refactor so that all vscode-specific infoview stuff is in `extension.ts`

EdAyers:
A refactoring, no new functionality.

This is to maximise code reuse with the lean-web-editor project.
All code that is specific to vscode is now in infoview/extension.ts.
So it should only be necessary to re-implement this file to get the infoview system
to work in lean-web-editor.

Additionally:

- index.tsx has been split in to Main.tsx and index.tsx. index.tsx is
  vscode specific
- Main has been split in to `Main` and `InfoView` components. This is to
  separate the stateful parts of Main which lean-web-editor will likely not use.
- onPin is now optional, if onPin is not provided then Info will not
  show the pin button
- rename DOM id `react_root` to `infoview_root` since lean-web-editor uses React and so react_root is no longer the root

## todo

- [ ] test etc
- [ ] get lean-web-editor working using the contents of infoview with only 'extension.ts' reimplemented. So likely I will have to change this again


infoview/extension.ts
```diff
@@ -1,50 +1,86 @@
+/* This file contains everything that is specific to the vscode extension
+implementation of the infoview. So the idea is that lean-web-editor
+shares the rest of this infoview directory with this project. */
+
+import * as trythis from '../src/trythis';
+export {trythis};
```
gebner, line 6: You could also move the regexps in the `trythis` module to the infoview (or shared) directory and then import them the other way around from the `trythis.ts` file.

I'd like to avoid multiple copies of the regexes if possible.

EdAyers: I don't quite get what you mean by this, but I don't understand what 'try this' is doing either.

gebner: If you call `squeeze_simp`, then it will print `Try this: simp only [foo]`.  You can then click on the "Try this" in the infoview and it will insert the suggestion in the editor.  We also have a code action for this, which is contained in `tacticsuggestions.ts`.  The `trythis.ts` module contains the regexps used to parse the suggestion.

Right now, the infoview imports the regexps from `src/trythis.ts`.  My proposal is to move `trythis.ts` to `infoview/trythis.ts` and have `tacticsuggestions.ts` import `infoview/trythis.ts` instead.  Then we would only require one copy of this file.

EdAyers: There's only one copy of the file at the moment though.

gebner: I assumed that you would copy the file to the web editor.

EdAyers: ah I see what you mean now

PR #207, opened 2020-07-12
feat(Try this): cancel 'by exact'

semorrison:
If there's a `by ` immediately before the tactic invocation, and the "Try this:" suggestion begins with `exact `, cancel the `by exact `.

### DIFF
--- a/src/tacticsuggestions.ts
+++ b/src/tacticsuggestions.ts
@@ -115,13 +115,24 @@ export class TacticSuggestions implements Disposable, CodeActionProvider {
         const lastPos = new Position(endLine, endCol)
         textEditor.selection = new Selection(lastPos, lastPos)
 
-        // Now check for `by ` before the tactic call, 
-        // and `exact ` at the beginning of the suggestion
+        // Now check for `exact ` at the beginning of the suggestion 
+        // and `by ` before the tactic call (or `begin ` before, and `, end` after)
         // and if both occur, remove them.
         if (suggestion.startsWith('exact ')) {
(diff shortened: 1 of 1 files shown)


src/tacticsuggestions.ts
```diff
@@ -72,7 +72,7 @@ export class TacticSuggestions implements Disposable, CodeActionProvider {
 
         // Start of the tactic call to replace
         const startLine = m.pos_line - 1;
-        const startCol = m.pos_col;
+        var startCol = m.pos_col;
```
bryangingechen, line 75:
```suggestion
        let startCol = m.pos_col;
```

semorrison: I'll take your word for it. :-)

src/tacticsuggestions.ts
```diff
@@ -115,6 +115,16 @@ export class TacticSuggestions implements Disposable, CodeActionProvider {
         const lastPos = new Position(endLine, endCol)
         textEditor.selection = new Selection(lastPos, lastPos)
 
+        // Now check for `by ` before the tactic call,
```
bryangingechen, line 118:
```suggestion
        // Now check for `by ` before the tactic call,
```
This should fix the linter complaint.

semorrison:
I now also look for `begin ___, end`.

PatrickMassot:
Having this working and merged before the workshop would be really great since we'll use `library_search` a lot during this week.

gebner:
Thanks!

semorrison:
Fantastic --- just restarted VSCode and it works!

PR #210, opened 2020-07-20
Remove the workaround for a now-fixed vscode issue

eric-wieser:
https://github.com/Microsoft/vscode/issues/14221 was fixed in late 2017


gebner:
Thanks!

PR #211, opened 2020-07-20
Make begin / end an autoclosing pair

eric-wieser:


utensil:
It seems to be better to add things like this as snippets like https://github.com/James-Yu/LaTeX-Workshop/blob/master/snippets/latex.json .

@eric-wieser Have you tested this locally? Is any of these working:

- input `begin` then the line becomes `beginend`
- the cursor stay between `begin` and `end`
- hit `Enter`, `begin` and `end` have the same identation

(the above are trivial, but I don't know if it still works combining the following:)

- input `be`, hit `tab`, then it's auto-completed to `beginend`
- then the rest of the above still works.

eric-wieser:
No, I'm afraid I have not tested this as I have not yet learnt how to.

eric-wieser:
Although it behaves exactly as typing `/-` does today.

I think you're right, this makes more sense as a snippet. Opened the weaker #213 instead.

eric-wieser:
xref https://github.com/microsoft/vscode/issues/583

gebner:
Sorry for not looking at this earlier!  I wasn't sure how comfortable this would be in practice so I wanted to try it out first.  But apparently there are vscode issues anyhow. :disappointed:

PR #213, opened 2020-07-27
Consider begin / end as brackets

eric-wieser:
Done for consistency with the vs-code verilog package, which also considers these as brackets.

https://github.com/mshr-h/vscode-verilog-hdl-support/blob/master/verilog.configuration.json#L13

This results in highlighting of matching begin/end:

image

Better yet, it also results in auto-indentation after `begin`


gebner:
Thanks!

PR #214, opened 2020-07-27
Add some basic snippets

eric-wieser:
Not sure if these are really all that useful, but though I'd suggest them anyway.

### DIFF
--- a/snippets/lean.json
+++ b/snippets/lean.json
@@ -1,0 +1,48 @@
+{
+    "lemma": {
+        "prefix": "lemma",
+        "body": [
+            "lemma ${1:name (args)} : ${2:sorry} := ${3:sorry}"
+        ],
+        "description": "Insert a one-line lemma"
+    },
+    "theorem": {
+        "prefix": "theorem",
+        "body": [
+            "theorem ${1:name (args)} : ${2:sorry} := ${3:sorry}"
+        ],
+        "description": "Insert a one-line lemma"
+    },
+    "example": {
+        "prefix": "example",
+        "body": [
+            "example : ${2:sorry} := ${2:sorry}"
+        ],
+        "description": "Insert a one-line example"
+    },
+	"namespace": {
+        "prefix": "namespace",
+        "body": [
+            "namespace ${1:name}",
+            "$0",
+            "end ${1}"
+        ],
+        "description": "Insert a namespace"
+    },
+	"structure": {
+        "prefix": "structure",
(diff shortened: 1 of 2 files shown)


gebner:
I think these could prove to be fairly controversial.  You automatically get a lot of text when you type `structure`.  I don't see an easy way to avoid the snippets since you need to type `structure` as a keyword.  Can we add an options in the settings to enable them?

Note that it's also possible to add the snippets to e.g. mathlib.

eric-wieser:
> I don't see an easy way to avoid the snippets since you need to type `structure` as a keyword. Can we add an options in the settings to enable them?

vscode lets you pick between the two options, and uses your last choice as a default:

image

eric-wieser:
> Can we add an options in the settings to enable them?

We could, but there are already extensions to manage unwanted snippets like https://github.com/svipas/vscode-control-snippets. https://github.com/Microsoft/vscode/issues/10565 is related.

gebner:
Ok, so I've tried them out a bit.

- They are not triggered automatically unless you explicitly select them in the dropdown.  This is good.
- The `structure` snippet does not follow the mathlib style guide.  The `:=` should be at the end of the first line.  The fields should not be indented.
- The input abbreviation does not work in snippets.  That is, you can't type `\nat` in a snippet field. :sob: 
- I'm not sure how much they help.  The lean syntax does not have much boilerplate.  At best I don't have to type `:=` but that's it.

I've posted a thread in the PR reviews stream on zulip.

snippets/lean.json
```diff
@@ -0,0 +1,48 @@
...
+        "description": "Insert a namespace"
+    },
+	"structure": {
+        "prefix": "structure",
+        "body": [
+            "structure ${1:name} ${2:(args : sorry)}",
+            ":=",
+            "\t${3:(field : sorry)}",
+            "",
+            "namespace ${1:}",
```
utensil, line 39: The lines below are a little too much. `structure` should not make too much assumption on the organization after the structure. If this snippet is called `theory` for boilerplates of developing a theory, maybe it would be OK but there would be much more.

snippets/lean.json
```diff
@@ -0,0 +1,48 @@
...
+            "variables ${2:}",
```
utensil, line 40: Usually variables should be declared in a named section.

bryangingechen: Declaring variables in a namespace is fine too.

Issue #208, opened 2020-07-12
Infoview doesn't work with Docker


arh4uwe:
My group uses a virtual machine called Docker when working on our projects. Basically, Docker creates a container to which VS Code can attach.

The issue is that while the Infoview panel works fine when the container is running, it goes completely blank when VS Code is attached to a Docker container. This issue didn't happen at all back when Lean Infoview was Lean Goal, so I think downgrading Lean, this plugin, or VS Code might be a temporary solution. How would I go about doing that?

bryangingechen:
You could try the [.vsix package](https://code.visualstudio.com/docs/editor/extension-gallery#_install-from-a-vsix) here: https://github.com/leanprover/vscode-lean/releases/tag/v0.15.15

bryangingechen:
[Related discussion on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/vscode.20infoview.20overhaul/near/203024285) (re: SSH-remote VS Code.)

arh4uwe:
Downgrading to an earlier version (0.15.15) made the issue go away. Hopefully Lean Infoview gets fixed in a later update. Until then, 0.15.15 will work for me.

sjkillen:
This is still an issue in 0.16.5 using a vscode remote container
Installing the .vsix @ 0.15.15 works for now
Note that the suggested method of .vsix install:
`code --install-extension myextension.vsix`
Does not work in vscode remote. Must install from GUI in extensions tab

bryangingechen:
Yeah, sadly I haven't had any time to look further into this at all. Feel free to post here or on Zulip if you have any ideas as to how we can get this working again.

sjkillen:
The culprit seems to be seems to be getMediaPath from infoview.ts:346
Currently this function is
```ts
private getMediaPath(mediaFile: string): string {
        // workaround for https://github.com/microsoft/vscode/issues/89038
        return this.staticServer.mkUri(join(this.context.extensionPath, 'media', mediaFile));
}
```
If I change it back to what it was in 0.15.5
```ts
private getMediaPath(mediaFile: string): string {
        // workaround for https://github.com/microsoft/vscode/issues/89038
        return Uri.file(this.context.asAbsolutePath(join('media', mediaFile)))
            .with({scheme: 'vscode-resource'}).toString();
}
```
The extension works in vscode docker
Reproduction:
build vsix with above change
clone https://github.com/sjkillen/example, open in container and install vsix

[...]

PR #217, opened 2020-08-05
Revert infoview:getMediaPath

sjkillen:
Revert getMediaPath method to the one used in v0.15.5
This allows the info view to work in vscode remote fixing #208

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1596,6 +1596,13 @@
 				"semver": "^5.5.0",
 				"shebang-command": "^1.2.0",
 				"which": "^1.2.9"
+			},
+			"dependencies": {
+				"semver": {
+					"version": "5.7.1",
+					"resolved": "https://registry.npmjs.org/semver/-/semver-5.7.1.tgz",
+					"integrity": "sha512-sauaDf/PZdVgrLTNYHRtpXa1iRiKcaebiKQ1BJdpQlWH2lCvexQdX55snPFyK7QzpudqbCI0qXFfOasHdyNDGQ=="
+				}
 			}
 		},
 		"crypto-browserify": {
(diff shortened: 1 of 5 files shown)


bryangingechen:
Rather than just revert this one instance, I would look at all the changes in https://github.com/leanprover/vscode-lean/commit/3b60fae4d95079beae2701a1f9a12995367dcece#diff-26aa7a3396b0249064623d10481785a2 (except for the change to `package_lock.json`) and revert those as well.

sjkillen:
Are there tests to be run in the launch tests task? When I launch the task it complains about `/out/test` being missing but the extension seems to work when I use it

bryangingechen:
I've never actually tried running the tests myself, so I'm fairly certain the answer is no. There do seem to be some tests in the `test/` directory but they don't look particularly useful. 

Regarding the complaint from the task, from what I can tell, `out/test/` is where any output files should end up, not where the tests are / would be located, so probably https://github.com/leanprover/vscode-lean/blob/2fd4cab7ed3ff0dffaeb21e9d5cd6eff9530e0fe/.vscode/launch.json#L21 should be changed so that `extensionTestsPath` points to `${workspaceRoot}/test` instead.

gebner:
I'd feel infinitely more comfortable with this change if we only use `asWebviewUri` for vscode versions where it has been fixed.

There are indeed no tests for the extension.  Even worse, the `asWebviewUri` bug is non-deterministic.  I could never reproduce it using `F5`, only when installing the extension regularly.

sjkillen:
I've added a switch to resort back to staticserver if the vscode version is less than 1.46.0 (https://github.com/leanprover/vscode-lean/issues/162#issuecomment-630549016)

I've only been testing outside the extension development host as I'm not sure it works for VScode remote

gebner:
I've bumped the required vscode version to 1.47 because I still got the unknown url scheme error on 1.46.1.  I think they delayed the fix by one version.

gebner:
Reference for delay: https://github.com/microsoft/vscode/issues/89038#issuecomment-637161257

PR #219, opened 2020-08-11
Add support for Unicode input with multiple cursors

### LINKED ISSUE #160: Multi-cursor unicode support
Unicode entry via backslash commands does not work in multi-cursor mode.

To recreate, make multiple cursors via cmd+option+down_arrow and type \Gamma. Hitting tab or space does not trigger the unicode character translation. Another symptom of this issue is that \Gamma is usually underlined when tabbing will trigger the unicode translation, but is not underlined any of the multiple cursors.

alexpeattie:
Add support for multiple cursors for Unicode input, closes #160 🙂. I've tested the following cases with single and multiple cursors:

- Using <kbd>Tab</kbd> key
- Using <kbd>Return</kbd> key
- Moving outside of range to be converted (e.g. with mouse or arrow keys)
- Typing a space
- Typing another <kbd>\\</kbd>

To accomplish multiple cursor support I replace `TextEditorAbbrevHandler.range` with `TextEditorAbbrevHandler.ranges` (any array of `Range`s) - otherwise the logic remains mostly the same.

#### Notes

- ~~In the current version there's slightly odd behaviour where if you type <kbd>\\\\</kbd> it get collapsed into <kbd>\\</kbd> (but only once, additional <kbd>\\</kbd>s appear as normal). This was due to the lines:~~

  ```js
  if (change.text === this.leader && this.rangeSize === 1) {
      await this.updateRange();
      return this.editor.edit((builder) =>
          builder.delete(new Range(change.range.start, change.range.end.translate(0, 1))));
  }
  ```

  ~~I'm not sure why this would be intended behaviour, so I've scrapped it for now.~~ **Edit:** Old behaviour restored

- `convertRange()` (now `convertRanges()`) currently wraps the replacement operation in a timeout with the comment:

  ```js
  // Without the timeout hack, inserting `\delta ` at the beginning of an
  // existing line would leave the cursor four characters too far right.
  ```

  I haven't been able to replicate this, so I've ditched the timeout.

#### Caveats

- "Hacky replacements" (⦃⦄, ⟦⟧ etc.) work with multiple cursors. However, getting all the cursors positioned inside the brackets after insertion would I think be a bit of a headache - so right now you just end up with a single cursor in the first pair of brackets:

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -1,6 +1,6 @@
-import { CancellationToken, commands, Disposable, DocumentFilter, Hover,
+import { CancellationToken, commands, Disposable, Hover,
     HoverProvider, languages, Position, Range, Selection, TextDocument,
-    TextDocumentChangeEvent, TextEditor, TextEditorDecorationType,
+    TextDocumentChangeEvent, TextDocumentContentChangeEvent, TextEditor, TextEditorDecorationType,
     TextEditorSelectionChangeEvent, window, workspace } from 'vscode';
 
 export interface Translations { [abbrev: string]: string | null }
@@ -80,7 +80,7 @@ export class LeanInputExplanationHover implements HoverProvider, Disposable {
 }
 /* Each editor has their own abbreviation handler. */
 class TextEditorAbbrevHandler {
-    ranges: (Range | undefined)[] = [];
+    ranges: Range[] = [];
 
     constructor(public editor: TextEditor, private abbreviator: LeanInputAbbreviator) {}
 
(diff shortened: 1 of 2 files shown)


gebner:
> In the current version there's slightly odd behaviour where if you type \\ it get collapsed into \ (but only once, additional \s appear as normal).

This is very much intentional because it allows you to type `"\r\n"`, etc.

> I haven't been able to replicate this, so I've ditched the timeout.

Ok, then this bug has probably been fixed in vscode in the meantime.

src/input.ts
```diff
@@ -249,7 +249,7 @@ export class LeanInputAbbreviator {
 
     get active(): boolean {
         const handler = this.handlers.get(window.activeTextEditor);
-        return handler && !!handler.range;
+        return handler && (handler.ranges.length > 0);
```
gebner, line 252: Why not `!!handler.ranges`?  It's eleven characters shorter!  Just kidding, but javascript semantics are fun.

src/input.ts
```diff
@@ -80,14 +80,14 @@ export class LeanInputExplanationHover implements HoverProvider, Disposable {
 }
 /* Each editor has their own abbreviation handler. */
 class TextEditorAbbrevHandler {
-    range: Range | undefined;
+    ranges: (Range | undefined)[] = [];
```
gebner, line 83:
```suggestion
    ranges: Range[] = [];
```

If I understand the code correctly.

alexpeattie: > This is very much intentional because it allows you to type `"\r\n"`, etc.

Ah OK, that makes sense - I'll have a go at reenabling that behaviour ☺️

alexpeattie:
> This is very much intentional because it allows you to type `"\r\n"`, etc.

Ah OK that makes sense, I'll have a go at reenabling that behaviour ☺️

src/input.ts
alexpeattie: 👍

gebner:
Thanks!

PR #223, opened 2020-08-23
escape quote in wordPattern

### LINKED ISSUE #222: Errors parsing language-configuration.json
While investigating why Bracket Colorizer 2 no longer works with Lean files, I noticed that even with just vscode-lean enabled, the following error appears in VS Code's dev tools:
```
Errors parsing file:///Users/chb/.vscode/extensions/jroesch.lean-0.16.10/language-configuration.json: [2721, 1] Invalid symbol
[2722, 1] Invalid symbol
[2723, 4] Invalid symbol
[2727, 1] Property name expected
[2728, 2] Invalid symbol
[2730, 1] 
[2732, 1] Value expected
(anonymous) @ languageConfigurationExtensionPoint.ts:105
```
(The error is being thrown [here](https://github.com/microsoft/vscode/blob/f74e473238aca7b79c08be761d99a0232838ca4c/src/vs/workbench/contrib/codeEditor/browser/languageConfigurationExtensionPoint.ts#L105) in VS Code.)

I can't figure out how to translate the `[2722, 1]` to actual positions in our configuration file, but maybe the fancy Unicode symbols aren't supported?

bryangingechen:
Fixes #222.


bryangingechen:
This seems to get Bracket Colorizer 2 working again too.

PR #225, opened 2020-08-28
Add support for katex.

EdAyers:
The design is that now if the user uses the special tags `InlineMath` or `BlockMath`, as in;
```
#html (h "BlockMath" [attr.val "math" "f(x) = \\int_{-\\infty}^\\infty \\hat f(\\xi)\\,e^{2 \\pi i \\xi x} \\,d\\xi"] [])
```
Then vscode will catch it and turn it in to a component. The component list is stored in a javascript object called `tagComponentDict`.
This does make the lean <-> vscode interface more complicated, because now lean is assuming that the widget renderer knows to render InlineMath and BlockMath as katex components.
So for example if someone wanted to write thier own widget renderer, there are now more assumptions that they need to be aware of.
However in the case of some key libraries like D3 and Katex I think this is not too great a sacrifice since the only widgets client at the moment is the vscode extension. Also adding this feature doesn't require any changes in core or break existing widgets in any way.

It would be nice at some point to have Lean be able to programmatically say which libraries it wants, but for now this will do.
Alternatively, at some point we should standardise which libraries and CSS classes lean expects to be available.

~~The style sheet for katex is loaded from a CDN instead of using webpack, because it was easier and less bloaty than getting webpack to include the font files.~~

### DIFF
--- a/src/infoview.ts
+++ b/src/infoview.ts
@@ -362,7 +362,6 @@ export class InfoProvider implements Disposable {
                 <meta http-equiv="Content-type" content="text/html;charset=utf-8">
                 <title>Infoview</title>
                 <style>${this.stylesheet}</style>
-                <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.12.0/dist/katex.min.css" integrity="sha384-AfEj0r4/OFrOo5t7NnNe46zW/tFgW6x/bCJG8FqQCEo3+Aro6EYUG4+cU+KJWu/X" crossorigin="anonymous">
             </head>
             <body>
                 <div id="react_root"></div>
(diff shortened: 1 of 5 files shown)


src/infoview.ts
```diff
@@ -362,6 +362,7 @@ export class InfoProvider implements Disposable {
                 <meta http-equiv="Content-type" content="text/html;charset=utf-8">
                 <title>Infoview</title>
                 <style>${this.stylesheet}</style>
+                <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.12.0/dist/katex.min.css" integrity="sha384-AfEj0r4/OFrOo5t7NnNe46zW/tFgW6x/bCJG8FqQCEo3+Aro6EYUG4+cU+KJWu/X" crossorigin="anonymous">
```
gebner, line 365: Since we already install `katex` via npm, could we also use it here instead of a CDN?

PR #226, opened 2020-09-15
remove leading and trailing whitespace from suggestions

bryangingechen:
This PR should improve support for multi-line suggestions. However, we might want to make more extensive changes to the suggestion format as well, e.g. wrap the suggestions in HTML tags so they look nicer in widget mode.

See https://github.com/leanprover-community/mathlib/pull/4093#issuecomment-692202104.

### DIFF
--- a/src/trythis.ts
+++ b/src/trythis.ts
@@ -1,5 +1,5 @@
 // Match everything after "Try this" until the next unindented line
-export const magicWord = 'Try this: ';
+export const magicWord = 'Try this:';
 export const regex = '^' + magicWord + '((.*\n )*.*)$';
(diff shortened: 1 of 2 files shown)


bryangingechen:
cc: @abentkamp

src/trythis.ts
```diff
@@ -1,5 +1,5 @@
 // Match everything after "Try this" until the next unindented line
-export const magicWord = 'Try this: ';
+export const magicWord = 'Try this:';
```
bryangingechen, line 2: Removing this space is necessary for the suggestions in https://github.com/leanprover-community/mathlib/pull/4093 to trigger.

abentkamp:
Looks good!

PR #234, opened 2020-11-15
Add widget effect support for specifying location of text to insert.

EdAyers:
See https://github.com/leanprover-community/lean/pull/501
and https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/paste.20the.20goal.20at.20the.20cursor

This commit allows `insert_text` effects to include position information. An `insert_text` effect now has two extra fields:

```
{ "kind": "insert_text"
, "line": number // the line number to insert at, can be negative for relative line_type
, "column"?: number
, "file_name"?: string
, "insert_type": "absolute" | "relative" // whether line property is relative to cursor or absolute
, "text": string
}
```

### DIFF
--- a/src/infoview.ts
+++ b/src/infoview.ts
@@ -215,7 +215,7 @@ export class InfoProvider implements Disposable {
         this.proxyConnection.send(JSON.parse(message.payload));
     }
     private async handleInsertText(message: InsertTextMessage) {
-        const new_command = message.text;
+        let new_command = message.text;
         let editor: TextEditor = null;
         if (message.loc) {
            editor = window.visibleTextEditors.find(e => e.document.fileName === message.loc.file_name);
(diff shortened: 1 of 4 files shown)


src/infoview.ts
```diff
@@ -232,14 +232,10 @@ export class InfoProvider implements Disposable {
...
-        // [hack] for now, we assume that there is only ever one command per line
-        // and that the command should be inserted on the line above this one.
-
+        new_command = new_command.replace(/\n/g, '\n' + margin_str);
+        new_command = `\n${margin_str}${new_command}`;
         await editor.edit((builder) => {
-            builder.insert(
-                prev_line.range.end,
-                `\n${margin_str}${new_command}`);
+            builder.insert(prev_line.range.end, new_command);
```
bryangingechen, line 238: This still only allows for text to be inserted on the line above the current one, right? For Johan's use case in # I think we want to be able to insert text right at the cursor.

EdAyers: ah I see

EdAyers: sorry I completely missed that

PR #236, opened 2020-12-04
Use two space indentation by default for lean code

eric-wieser:


package.json
```diff
@@ -394,7 +394,12 @@
 					"group": "navigation@2"
 				}
 			]
-		}
+		},
+		"configurationDefaults": {
+			"[lean]": {
+				"editor.tabSize": 2
+			}
+		},
```
eric-wieser, line 402:
```suggestion
		}
```

PR #238, opened 2020-12-16
Render markdown descriptions in settings

eric-wieser:
This is described here: https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema

I have not tested this.


gebner:
LGTM, I'll merge it after you've tried it out.

eric-wieser:
Ok, I tried this out:

image

image

Seems to work fine

Issue #239, opened 2020-12-29
Refactoring Of Abbreviation Feature


hediet:
Currently, the abbreviation feature (hover + completion) is all implemented in `input.ts`.

I think it might make sense to clean up that file a little bit and to split it up.
I'm working on a PR right now that presents some ideas. This issue can be used to discuss stuff before I submit the PR in the next few days.

What do you think?

After that restructuring, I want to improve the abbreviation feature to give something back for all the lean-help I got on zulip.
Ideas include:
* Autocompletion for abbreviations
* Live preview of the abbreviated symbol

hediet:
Also, the current range tracking algorithm has some issues with multi cursor support:

![ee62cf9a-b7c4-4624-a034-d84c8ca4685d](https://user-images.githubusercontent.com/2931520/103302694-806ec600-4a04-11eb-8649-5f421fc7b601.gif)

bryangingechen:
Multi-cursor support for unicode translations was added somewhat recently in #219 by @alexpeattie.

hediet:
I guess the current implementation assumes that tracked ranges don't get shifted.
However, this can actually happen when multiple ranges are in a single line.
Thanks for linking the PR!

alexpeattie:
Yes I think a clean refactor of this would be great @hediet 🙂! I think it would be great to do away with all of the "hacky replacement" stuff if possible, and there are other suboptimal things in there, like needing separate handlers for `convert` and `convertWithNewline`

hediet:
I think it would be nice if
1) abbreviations could be replaced as soon as they are not the prefix of any longer abbreviation anymore (this would be a generalization of extending a range with a whitespace character). This could also solve the `\\` case with `\` being an abbreviation for `\`.
2) abbreviations can have cursor placeholders where the cursor is placed after the abbreviation is unfolded. This would fix the hacky stuff.

hediet:
What are the usecases for `convert` and `convertWithNewline`? Why wouldn't you just press enter or the arrow key?

alexpeattie:
Tab and Enter are bound to `convert` and `convertWithNewline` respectively, where the latter differs in that it inserts a `\n`. The arrow keys don't use either, but rather the `onSelectionChanged` event ☺️

hediet:
But why isn't convert and convertWithNewline handled by a text change event?

alexpeattie:
[...]

PR #240, opened 2020-12-30
Rewrite of the abbreviation feature.

hediet:
1) Fixes some bugs in multi-cursor scenarios.
2) Gets rid of several hacks by introducing a more general system that unifies space/enter and abbreviation-character handling.
3) Removes now obsolete command "lean.input.convertWithNewline".
4) Abbreviations are now replaced as soon as possible if the abbreviation is unique.
5) Adds mobx dependency and makes the abbreviation config observable.

See #239.

Smart replacing if abbreviation is unique:
eager-replacing

Multi-cursor demo:
multi-cursor

Except (4) and the bug fixes, no semantics should be changed with this PR.


.eslintrc.js
```diff
@@ -116,7 +116,7 @@ module.exports = {
         "no-throw-literal": "error",
         "no-trailing-spaces": "error",
         "no-undef-init": "error",
-        "no-underscore-dangle": "error",
+        "no-underscore-dangle": "off",
```
hediet, line 119: This setting is turned off to allow fields that can be mutated privately (`_foo: string`) but that are readonly publically (`get foo(): string { return this._foo; }`).

package.json
```diff
@@ -339,13 +339,7 @@
...
 				"key": "tab",
 				"mac": "tab",
-				"when": "editorTextFocus && editorLangId == lean && lean.input.isActive"
-			},
-			{
-				"command": "lean.input.convertWithNewline",
-				"key": "enter",
-				"mac": "enter",
-				"when": "editorTextFocus && editorLangId == lean && lean.input.isActive && !suggestWidgetVisible && !renameInputVisible && !inSnippetMode && !quickFixWidgetVisible && (!vim.active || vim.mode == 'Insert')"
+				"when": "editorTextFocus && !suggestWidgetVisible && editorLangId == lean && lean.input.isActive"
```
bryangingechen, line 342: Can you update the README to remove `lean.input.convertWithNewline` as well?

Maybe you could even add the GIF showing off the abbreviation from your PR comment.

hediet: > Can you update the README to remove lean.input.convertWithNewline as well?

This is currently not mentioned in the readme - so there is nothing to remove.

> Maybe you could even add the GIF showing off the abbreviation from your PR comment.

I would wait until I experimented with the intellisense feature. Also, this GIF is not really meant to document the feature - too much is going on in it.

bryangingechen: Whoops, I thought I'd added all options to the README when I revamped it, but I guess I skipped this one. Sorry for the noise!

hediet:
There is still a cursor positioning bug - it happens when typing `\ax`. Will fix it soon.

src/features/abbreviation/abbreviations.json
```diff
@@ -0,0 +1,1753 @@
...
+	"{{}}": "⦃$CURSOR⦄",
+	"[]": "[$CURSOR]",
+	"[]_": "[$CURSOR]_",
+	"[[]]": "⟦$CURSOR⟧",
+	"<>": "⟨$CURSOR⟩",
+	"()": "($CURSOR)",
+	"()_": "($CURSOR)_",
+	"([])'": "⟮$CURSOR⟯",
+	"f<>": "‹$CURSOR›",
+	"f<<>>": "«$CURSOR»",
```
ocfnash, line 13: Given https://github.com/leanprover/vscode-lean/pull/249 should we do something like this?
```suggestion
	"f<<>>": "«$CURSOR»",
	"[--]": "⁅$CURSOR⁆",
```

hediet: Yes! However, I think this should be added as resolution in the merge commit.

hediet:
> There is still a cursor positioning bug - it happens when typing `\ax`. Will fix it soon.

Fixed!

src/features/abbreviation/abbreviations.json
hediet: I just rebased my changes.

alexjbest:
Sorry, I haven't tested this myself, but it looks like this changes the behaviour without an option to revert to the original behaviour is that right?

Personally I like the sound of the changes, but I'm a bit worried that we should at least allow people to use the old system if they already have the muscle memory to press tab/enter at the appropriate moment and dont want to change.

hediet:
I will add a setting for that which can disable this behavior!

gebner:
Thanks for putting effort into improving the input mode!  Some comments:

The new directory structure is very "enterprisey".
 - Please keep the `translations.json` file where it is right now, in the root directory.  Other projects rely on this location.
 - Also, please avoid deeply-nested directory structures like `src/features/abbreviation/AbbreviationRewriterFeature/AbbreviationRewriterFeature.ts`.  Please make just a single directory `src/abbreviation` and put everything in there.

Also feel free to put more functions/classes in a single file, this is not java.

Please run `eslint` on the files (I think there's something wrong with `npm run lint` though, because it doesn't lint the new files...).

What is the reason to add a new `Range` type (i.e. why is `vscode.Range` not enough)?

Finally, please add the setting that Alex mentioned.

hediet:
> What is the reason to add a new Range type (i.e. why is vscode.Range not enough)?

Like the docs say: `Range` is offset/length based in contrast to `vscode.Range` which is line/column based.

It is way easier to get correctness with offset based datastructures, as you would need to detect line breaks in text that got inserted.

> Please keep the translations.json file where it is right now, in the root directory. Other projects rely on this location.

Can do that, but I still find it confusing that it is called `translations.json`.

> Also, please avoid deeply-nested directory structures like src/features/abbreviation/AbbreviationRewriterFeature/AbbreviationRewriterFeature.ts. Please make just a single directory src/abbreviation and put everything in there.

I also can do that, but personally, I'm against it. The "abbreviation" feature consists of several subfeatures - the rewriting, the hover-provider and maybe someday an autocompletion. The rewrite feature has these four entities (`AbbreviationRewriter` which is per editor, `AbbreviationRewriterFeature` which is the thing that instantiates all the rewriters, `Range` which is a util thing and `TrackedAbbreviation` which is per tracked abbreviation in an editor) - I think they deserve a separate folder.

> Also feel free to put more functions/classes in a single file, this is not java.

I used to put many things into single files as it has been done in `input.ts`. But I find it very hard to get an overview of `input.ts`, as I need to scroll over all these details to figure out the larger picture.

> Please run eslint on the files (I think there's something wrong with npm run lint though, because it doesn't lint the new files...).

This is strange, I can try to fix that. I also highly recommend to setup prettier for formatting, enable strict mode and lift some eslint/tslint rules. Currently, tslint is doing stuff that other tools can do better.

> Finally, please add the setting that Alex mentioned.

Will do that.

gebner:
> It is way easier to get correctness with offset based datastructures, as you would need to detect line breaks in text that got inserted. This is the reason for some bugs in the current implementation, if the multiple cursors are added to the same line.

This makes sense!

> Also, if other projects depend on it, this should be documented. I added some <> -> <$CURSOR> abbreviations where the cursor is moved to $CURSOR after rewriting. I guess other projects might not like that and thus this needs to be changed.

Good point!  Feel free to rename it to `abbreviations.json` then if the format changes.  But I'd still prefer to keep it in the root directory (where it is more visible).

> > Also, please avoid deeply-nested directory structures like src/features/abbreviation/AbbreviationRewriterFeature/AbbreviationRewriterFeature.ts.
> I also can do that, but personally, I'm against it. The "abbreviation" feature consists of several subfeatures - the rewriting, the hover-provider and maybe someday an autocompletion. The rewrite feature has these four entities [...] - I think they deserve a separate folder.

The overriding principle here is consistency with the rest of the codebase, which doesn't use directory hierarchies either.

And I'm not even against subdirectories.  I like `src/abbreviation`, since there are now a lot of files for this features and it's great that they're grouped together.  `src/abbreviation/rewriter/Range.ts` would still be ok.  But repeating the name three times in the path is just too much:
`src/features/abbreviation/AbbreviationRewriterFeature/AbbreviationRewriterFeature.ts`

The `features` directory is superfluous, it only contains a single entry.  The word "abbreviation" is repeated three times.  And the even longer "AbbreviationRewriterFeature" is repeated twice.

> This is strange, I can try to fix that. I also highly recommend to setup prettier for formatting, enable strict mode and lift some eslint/tslint rules. Currently, tslint is doing stuff that other tools can do better.

Feel free to add prettier if you'd like.  (But in a different PR please.)

hediet:
> The features directory is superfluous

It very much is (currently)! But I can only recommend separating feature code from core/infrastructure code.
For me, a feature is some clear unit that you could delete and after deletion, you only need to touch like 1-5 lines of code to make the entire project work again (without the feature).
`extension.ts`, `server.ts`, `shared.ts`, `util.ts`, ... are all no features, while `completion.ts`, `statusbar.ts`, `infoview.ts`, ... (I guess) are.
Organizing code into features makes it very easy to identify the code you need to touch for implementing a certain effect.
The current `src` dir has 25 files which makes linear search increasingly difficult. Putting a decision tree in front of those files might make that much easier.

I'll remove the `features` folder from this PR, but I can really recommend this structure.

> But repeating the name three times in the path is just too much:

I understand your point and I will change it, but there was some reasoning behind chosing those names. If you configure the indentation in VS Code of nested folders, long names are not problematic anymore.
My naming scheme (and the one we use where I work) is the following:
* `kebab-case-name(.ts)?` if the item refers to some general concept.
* `UpperCamelCase(.ts)?` if the item refers to a main symbol of that name.
The folder `AbbreviationRewriterFeature` refers to the class `AbbreviationRewriterFeature` as being the most important symbol in that folder.

> Feel free to add prettier if you'd like. (But in a different PR please.)

I will!

> But I'd still prefer to keep it in the root directory (where it is more visible).

There was also a technical change that required me to move it - [I changed the code to bundle the json file directly as source](https://github.com/leanprover/vscode-lean/pull/240/files#diff-a2df975a286316631aec466cca41b6cfe6f79848cc4da7527e97d1d30aafc28aR4), rather than [loading it as file dynamically](https://github.com/leanprover/vscode-lean/blob/master/src/extension.ts#L79).
My change has the advantage that the compilation process checks the file path (and requires no dependency to `load-json-file`), while the dynamic import does not do that.

However, if `abbreviations.json` is now treated as source file, typescript puts its restrictions on it:

> File 's:/dev/contrib/vscode-lean/abbreviations.json' is not under 'rootDir' 's:/dev/contrib/vscode-lean/src'. 'rootDir' is expected to contain all source files.ts(6059)

I can try to fix that, even though I guess I need to fallback to the mechanism used in the current master.

gebner:
> There was also a technical change that required me to move it - I changed the code to bundle the json file directly as source, rather than loading it as file dynamically.

Ah, if there's a good reason then that's fine too.  We have quite a few users though that contribute to the file but don't develop the extension, so it'd be great if it was still easy to find (from the github page).

> Organizing code into features makes it very easy to identify the code you need to touch for implementing a certain effect.
The current src dir has 25 files which makes linear search increasingly difficult. Putting a decision tree in front of those files might make that much easier.

I agree, and that's why I said that a new directory for the abbreviations feature is a great idea.

Different projects have different organization styles.  It is better to be consistent within a project than to conform (partially) to a style guide, no matter how well-thought out it may be.  The optimal organization also depends on the size of a project, a 10-line script requiring a different setup than a 100k line project.

I am sure that at your company all new repositories conform to these style guides, and this is great for consistency and uniformity, and makes it easier for new people to get started on a project.  But this is not your company, and following such an elaborate style guide can even be detrimental since nobody here is familiar with it, negating some of the advantages.  This holds doubly if the style is only followed partially, for a single feature.

hediet:
> This holds doubly if the style is only followed partially, for a single feature.

Very true and I will change these things in this PR to be more consistent with the existing code.

Nonetheless, I think with 25 files in the `src` folder on master, it might already make sense to organize them in feature-like folders.

hediet:
* I got rid of the features folder
* renamed the folders to a shorter name
* ran eslint
* added the setting key `lean.input.eagerReplacementEnabled` that can be used to turn off that feature (on by default).
* added some docs for `abbreviation.json`

hediet:
🎉

src/abbreviation/rewriter/AbbreviationRewriter.ts
```diff
@@ -0,0 +1,256 @@
...
+	): Promise<void> {
+		if (abbreviations.length === 0) {
+			return;
+		}
+		for (const a of abbreviations) {
+			this.trackedAbbreviations.delete(a);
+		}
+
+		// Wait for VS Code to trigger `onDidChangeTextEditorSelection`
+		await waitForNextTick();
```
eric-wieser, line 101: My understanding is that this timeout is non-deterministic, and is the cause of the "sometimes the replacement doesn't happen" bug. It happens more when lean is busy.

hediet: > "sometimes the replacement doesn't happen"

This happens if the user changes the text buffer before the replacment is committed. This is unavoidable, it can only be made less likely to happen (I only learned recently about this, sorry for not having taken this into account when I wrote that code).

Internally, VS Code has a version number per text document that is incremented on every edit. Edits are sent asynchronously to the extensions and if an extension issues an (async) text modification it is only applied if the version of the target document has not changed in the meantime.

`waitForNextTick` makes it more likely that a user can change the text buffer before the extension can commit the text modification.

However, `waitForNextTick` is also required to position the cursor correctly after the replacment has been committed.

Your fix is a good workaround, but maybe the extension can actually try a second time to perform this replacement. Sadly (afaik) VS Code does not allow transactional updates that involve text and cursor modification.

PR #241, opened 2020-12-30
Configure VS Code to start the watch task (rather than the build task) when debugging the extension.

hediet:
This requires [this extension](https://marketplace.visualstudio.com/items?itemName=eamodio.tsl-problem-matcher) though to detect when webpack has finished building the output.

See #242.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,7 +11,7 @@
             "stopOnEntry": false,
             "sourceMaps": true,
             "outFiles": [ "${workspaceRoot}/out/src/**/*.js" ],
-            "preLaunchTask": "${defaultBuildTask}"
+            "preLaunchTask": "watch"
         },
         {
             "name": "Launch Tests",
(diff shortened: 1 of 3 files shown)


gebner:
Can you explain what's better about watch as opposed to build?

hediet:
`webpack --watch ...` spawns a process that keeps running and never finishes.
At first, it behaves like `webpack ...` - it just builds the extension and the webview. After the first successful build however, it watches all files and builds again if a source file changes.
This incremental build is much faster than the initial build, which makes this ideal for development.
Also, you don't need to build the extension every time you change something - you can just reload the window of the VS Code debugee instance.

hediet:
In other words: Currently, if you click here:
image

It needs to rebuild the entire extension (and I guess it also rebuilds the webview). This takes 5-10 seconds.

This PR makes this a no-op (0 seconds), since changes are integrated incrementally as soon as they happen.

gebner:
Ah, that's great then!

PR #253, opened 2021-01-17
Compatibility with vscode-lean4

mhuisi:
This PR adds compatibility for the new version of [vscode-lean4](https://github.com/mhuisi/vscode-lean4) that was built on top of vscode-lean.

For both extensions not to trample on one another, the following changes have been made to vscode-lean and vscode-lean4:
1. Upon being activated, both extensions check the lean version and yield if the major version is not their expected one. To do this, instead of reading the `leanpkg.toml` as previously suggested, `lean --version` is called in the workspace folder. This ensures compatibility with the nix setup of Lean 4, where no `leanpkg.toml` file exists. If no workspace folders are open, `lean --version` is called in the process environment of the editor to look for a global version of Lean for users that want to edit single Lean 4 files. If we cannot find `lean`, vscode-lean4 displays an error and yields to vscode-lean; this ensures that the existing vscode-lean behavior is preserved.
2. Lean 3 remains associated with `*.lean` files by default, whereas vscode-lean4 sets the language ID of files dynamically if the version was detected to be Lean 4.
3. In vscode-lean4, all the Lean 3 specific files of vscode-lean were moved to a folder `old`, `package.json` was adjusted and `syntaxes/lean.json` was replaced with `syntaxes/lean4.json`.
4. vscode-lean4 still contributes a `lean` language stub without a corresponding configuration to get activated at all when vscode-lean is not installed and a `*.lean` file is opened. This does not conflict with the `lean` language contributed by vscode-lean.

After this PR is merged I will transfer the ownership of vscode-lean4 to leanprover-community. Then we can take down the published vscode-lean4 from my user account and publish it under the leanprover account.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
 	"name": "lean",
 	"displayName": "lean",
 	"description": "An extension for VS Code which provides support for the Lean language.",
-	"version": "0.16.21",
+	"version": "0.16.20",
 	"publisher": "jroesch",
 	"engines": {
 		"vscode": "^1.38.0"
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
 	"name": "lean",
 	"displayName": "lean",
 	"description": "An extension for VS Code which provides support for the Lean language.",
-	"version": "0.16.20",
+	"version": "0.16.21",
 	"publisher": "jroesch",
 	"engines": {
 		"vscode": "^1.38.0"
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -26,7 +26,7 @@ async function checkLean3(): Promise<boolean> {
     if (!lean4) {
         return true;
     }
-    return (await lean4.activate()).aborted;
+    return !(await lean4.activate()).isLean4Project;
 }
 
 // Seeing .olean files in the source tree is annoying, we should
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -41,14 +41,12 @@ async function checkLean3(): Promise<boolean> {
 		// Specifically, if the extension was not opened inside of a folder, it
         // looks for a global installation of Lean.
         // In vscode-lean4 we do this for single-file language server support,
-        // here we do it to give the LeanpkgService a chance to display its error
-        // when we already know that vscode-lean4 will not be activated.
+        // here we do it to use the same binary as vscode-lean4.
         const { stdout, stderr } = await promisify(exec)(cmd, {cwd: folderPath})
-		const filterVersion = /Lean \(version (\d+)\.(\d+)\.(\d+), commit [^,]+, \w+\)/
+		const filterVersion = /Lean \(version (\d+)\.(\d+)\.([^,]+), commit [^,]+, \w+\)/
 		const match = filterVersion.exec(stdout)
 		if (!match) {
-			void window.showErrorMessage(`'${cmd}' returned incorrect version string '${stdout}'.`)
-			return false
+			return true
         }
 		const major = match[1]
 		if (major !== '3') {
@@ -56,8 +54,7 @@ async function checkLean3(): Promise<boolean> {
         }
 		return true
 	} catch (err) {
-		void window.showErrorMessage(`Could not find Lean version by running '${cmd}'.`)
-		return false
+		return true
 	}
 }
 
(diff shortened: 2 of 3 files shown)


mhuisi:
One issue with this approach is perhaps that if `lean` is not installed (but elan is), the version regex will fail as the right version of Lean is being installed. Another issue is that both extensions launch the command, and hence there will be two concurrent invocations of elan trying to install Lean, which I'm not sure it can handle.

I might need to think about this a little more to make it foolproof before this can be merged!

package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "lean",
 	"displayName": "lean",
 	"description": "An extension for VS Code which provides support for the Lean language.",
-	"version": "0.16.20",
+	"version": "0.16.21",
```
gebner, line 5: Please do not increment the version yourself.  I'll do this when I make a release.

src/extension.ts
```diff
@@ -21,6 +23,41 @@ import { LeanTaskGutter, LeanTaskMessages } from './taskgutter';
 import { StaticServer } from './staticserver';
 import { LibraryNoteLinkProvider } from './librarynote';
 
+function executablePath(): string {
+    return workspace.getConfiguration('lean').get('lean.executablePath', 'lean')
+}
+
+async function checkLean3(): Promise<boolean> {
+    const folders = workspace.workspaceFolders
```
gebner, line 31: There is a lot of code in `src/server.ts` that detects whether we should offer to install elan, or whether we need to do some workarounds on windows.  This should probably be done here instead or shared with the server code.

mhuisi: I have a different proposal. What if `checkLean3` becomes the following:
```typescript
async function checkLean3(): Promise<boolean> {
    const lean4 = extensions.getExtension('jroesch.lean4');
    if (!lean4) {
        return true;
    }
    return (await lean4.activate()).aborted;
}
```
vscode-lean4 would provide an API for `aborted`, which is the result of a similar version check, but for Lean 4. I already tried this out and the effects are as follows:
- If vscode-lean4 is not installed, vscode-lean behaves exactly as before.
- If both are installed while the corresponding local toolchain is not and a Lean 3 project is opened, vscode-lean takes a little longer to activate while the elan invocation of vscode-lean4 downloads the corresponding Lean version. If it does not succeed at this for whatever reason, vscode-lean4 emits an error where it is visibly obvious that it was produced by vscode-lean4 and the behaviour of vscode-lean remains exactly as before.
- The changes to vscode-lean are minimal.
- My concern about concurrent invocations of elan from above is resolved.

Thoughts? If the longer activation time without visual cues is an issue, I could try to find a way to stream the elan stderr to the vscode output in vscode-lean4, similar to what vscode-lean does to show that work is being done.

package.json
mhuisi: Oops, sorry!

mhuisi:
> One issue with this approach is perhaps that if `lean` is not installed (but elan is), the version regex will fail as the right version of Lean is being installed.

As it turns out, the output while elan is installing stuff is emitted to stderr, not stdout, so the regex is fine.

src/extension.ts
gebner: This approach sounds great to me.  Most of the corner cases with the Lean 3 extensions are for newcomers, which hopefully don't use Lean 4 yet.

One thing: please don't use `jroesch.lean4` for the new extension.  Instead use `leanprover.lean4`.  If you tell me the email address you use for your azure account, then I can make you a co-owner of the leanprover publisher id.

src/extension.ts
```diff
@@ -21,6 +21,14 @@ import { LeanTaskGutter, LeanTaskMessages } from './taskgutter';
 import { StaticServer } from './staticserver';
 import { LibraryNoteLinkProvider } from './librarynote';
 
+async function checkLean3(): Promise<boolean> {
+    const lean4 = extensions.getExtension('leanprover.lean4');
+    if (!lean4) {
+        return true;
+    }
+    return (await lean4.activate()).aborted;
```
gebner, line 29: I really don't like the name `aborted` here, can we pick something more self-explanatory and concrete like `isLean3Project` or `didNotFindLean4` or ...?

PR #254, opened 2021-01-20
Don't replace empty abbreviations

hediet:


gebner:
Did you see https://github.com/leanprover/vscode-lean/commit/01184cf95c3221d14c5a3ab860af7c9ff9e5e2c9?

hediet:
Not until now. Your fix is actually nicer, thanks! I hope you didn't have to debug much.

PR #262, opened 2021-05-02
 Add four new submenus to the editor context menu

hmonroe:
The submenus are for existing Ctrl-Shift-P commands; and for leanproject, elan, and
lean commands. See Zulip discussion https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Simplifying.20Mathlib.20install.20without.20command.20line/near/236595136

### DIFF
--- a/package.json
+++ b/package.json
@@ -178,7 +178,7 @@
 			{
 				"command": "lean.restartServer",
 				"category": "Lean",
-				"title": "Restart",
+				"title": "Restart Lean Server",
 				"description": "Restart the Language Server."
 			},
 			{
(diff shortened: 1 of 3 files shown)


eric-wieser:
Can you paste a screenshot giving a feel for what these new menus look like?

hmonroe:
Sure, here are four screenshots. 
Lean Actions
Lean Project Commands
elan Commands
Lean Command Line

eric-wieser:
The [extension guidelines](https://code.visualstudio.com/api/references/extension-guidelines#context-menus) suggest that these commands do not belong in a context menu, as they are not "contextually appropriate" (they act on a global lean install, not on the file / line / character being right clicked on).

I also question whether most of these commands make any sense to trigger from the GUI. To name a few at random:
* `lean -h` - why would I want command line help if I'm not using the command line already?
* `elan run` - this just emits a message `error: The following required arguments were not provided` ... `Usage: elan.exe run [FLAGS] <toolchain> <command>...`
* Any of the commands that just `echo` a help string ('leanproject.new', `leanproject.pr`, etc)

We should only add things to the VSCode UI if we can think of a scenario when users would use them - adding every command line invocation without first checking to see if it makes sense to invoke from a GUI is not a good use of time or editor menu space!

hmonroe:
Fair points--I would prioritize finding a way for users to open the Tutorials/Theorem Proving in Lean/Mathematics in Lean from within VS Code rather than the command line. Here are some options: (1) the user clones the relevant repository from the generic "Clone" command on the welcome page, and VS Code offers to install mathlib; (2) create a menu with these three projects plus plus a dialog to create a new project or contribute to mathlib. As a lower priority, I could add Restart Lean Server or Delete Zombies (if that is even a priority) to the blue bar at the bottom, which already has some Lean options such as Check Visible Files. What do you think?

eric-wieser:
> I would prioritize finding a way for users to open the Tutorials/Theorem Proving in Lean/Mathematics in Lean from within VS Code rather than the command line.

We already have this:

image

Which opens:

image

eric-wieser:
> create a menu with these three projects plus plus a dialog to create a new project or contribute to mathlib

Perhaps we could take some inspiration from what the Python extension does:

image

hmonroe:
Beautiful! I see the source code is available is here so I will try to do that. https://github.com/microsoft/vscode-python/blob/3698950c97982f31bb9dbfc19c4cd8308acda284/src/startPage-ui/startPage/startPage.tsx

eric-wieser:
Before you go down that path too far (ie, making all the buttons work), I'd recommend making a list of items you think such a page should have, and getting some feedback on zulip.

hmonroe:
I will solicit input on Zulip, but my immediate thought is an HTML welcome page similar to the Python extension, with sections for: (1) Load tutorials, Theorem Proving in Lean, Mathematics in Lean, Logic and Proof, Hitchhikers Guide to Logical Verification, i.e. what is on this page https://leanprover-community.github.io/learn.html. (2) Create a new standalone project and push to github if desired. (3) Contribute to mathlib; and (4) reopen a recently project.

Technically this seems straightforward: creating a web page (the extension already has two, for info view and doc view), making the welcome page pop up on startup unless the user declines that, and keeping track of the user's preferred root directory while providing an option to locate repositories elsewhere.

Once the other pieces are in place, my ideal install process would be (1) install Github for Windows; VS Code; and the Lean extension, and then the extension should install elan and leanproject.exe if not already installed. And if the user clones a github repository for a Lean project not using the web page above, the extension should detect that and make sure elan and leanproject are run as needed.

Some other features missing from the extension are documentation for reserved keywords like "cons" and "theorem", code folding (by section, by definition/theorem, namespace), help filling in imports. Also, a better way is needed to ensure that Lean projects like the tutorials/books work with the latest version of lean--for instance, a Github workflow for the tutorials should detected that mathlib is updated and provide an alert if the tutorials no longer build. And the mathlib maintainers might be alerted if a mathlib release will break other projects. Isabelle and Mizar have systems in place to maintain consistency but mathlib does not seem to, in a new version of "DLL hell". elan should be phased out some day.

hmonroe:
Closing pull request

hmonroe:
Closing pull request

PR #265, opened 2021-05-12
Correctly highlight lean expressions inside `@[...]` and `attribute […]`

eric-wieser:
This moves a bunch of the existing patterns into a new "expressions" group, which can be re-entered.

image


gebner:
Thanks!

eric-wieser:
I think this might need a follow-up; presumably comments are also allowed within the `[]`?

eric-wieser:
Followed up with #266

PR #270, opened 2021-05-26
fix(syntaxes/lean): implicit binder brackets are forbidden in symbol names

eric-wieser:


syntaxes/lean.json
```diff
@@ -275,7 +276,7 @@
           "beginCaptures": {"0": {"name": "punctuation.definition.binder.semiimplicit.begin.lean"}},
           "endCaptures": {"0": {"name": "punctuation.definition.binder.semiimplicit.end.lean"}},
           "patterns" : [
-            { "begin": ":", "end": "(?=\\⦄)",
+            { "begin": ":", "end": "(?=⦄|\\})",
```
eric-wieser, line 279:
```suggestion
            { "begin": ":", "end": "(?=⦄|\\}\\})",
```

PR #281, opened 2021-10-12
Get position

javier-m:
Added the widget argument (containing line and column) to the callback

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -43,7 +43,7 @@ const LEAN_MODE: DocumentFilter = {
     // scheme: 'file',
 };
 
-export async function activate(context: ExtensionContext): Promise<void> {
+export async function activate(context: ExtensionContext) {
     const isLean3 = await checkLean3();
     if (!isLean3) {
         return;
@@ -107,9 +107,10 @@ export async function activate(context: ExtensionContext): Promise<void> {
     context.subscriptions.push(new LeanStatusBarItem(server, roiManager));
 
     let staticServer = null;
+    let infoView : InfoProvider = null;
     function waitStaticServer() {
         // Add info view: listing either the current goal state or a list of all error messages
-        const infoView = new InfoProvider(server, LEAN_MODE, context, staticServer);
+        infoView = new InfoProvider(server, LEAN_MODE, context, staticServer);
         context.subscriptions.push(infoView);
         context.subscriptions.push(new DocViewProvider(staticServer));
         // Tactic suggestions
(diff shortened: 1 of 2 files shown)


src/infoview.ts
```diff
@@ -108,6 +110,10 @@ export class InfoProvider implements Disposable {
         }
     }
 
+    onNewTacticState(handler_name: string, handler: (state: string, widget: object) => void): void {
```
gebner, line 113:
```suggestion
    onNewTacticState(handler_name: string, handler: (loc: Location, state: InfoResponse) => void): void {
```

The `widget` field doesn't really contain any information (it's an id that you pass to the get_widget command).  You probably also want the location (filename & position) so that you can insert something there.

Can you explain what the `handler_name` should be?  Is this the name of the other extension?

I'm not sure if it's possible, but the nicest way would be to use the `EventEmitter` class from vscode:
```typescript
    private onNewTacticStateEmitter = new EventEmitter<{ location: Location, info: InfoResponse }>();
    onNewTacticState = this.onNewTacticStateEmitter.event;
```

src/extension.ts
```diff
@@ -128,4 +129,6 @@ export async function activate(context: ExtensionContext): Promise<void> {
 
     context.subscriptions.push(languages.registerDocumentLinkProvider(LEAN_MODE,
         new LibraryNoteLinkProvider()));
+
+    return {'infoView': infoView}; // export infoView for other plugins to use
```
gebner, line 133:
```suggestion
    return {onNewTacticState: infoView.onNewTacticState}; // export infoView for other plugins to use
```

I'd rather export only the event.

src/infoview.ts
```diff
@@ -120,12 +126,17 @@ export class InfoProvider implements Disposable {
                     payload: JSON.stringify(e)
                 })
             ),
-            this.proxyConnection.jsonMessage.on(e =>
+            this.proxyConnection.jsonMessage.on(e => {
+                if ('record' in e && 'state' in e.record && 'widget' in e.record){
```
gebner, line 130: This will trigger on every hover as well (i.e., when you move the mouse around).  A better place to intercept would be here: https://github.com/leanprover/vscode-lean/blob/2b43982c4c6305a0f20f156152a60613a6f1a683/infoview/info.tsx#L173
You'd then need to send a message from the webview back to the extension like this: https://github.com/leanprover/vscode-lean/blob/2b43982c4c6305a0f20f156152a60613a6f1a683/infoview/server.ts#L25-L27
And receive the message here:
https://github.com/leanprover/vscode-lean/blob/2b43982c4c6305a0f20f156152a60613a6f1a683/src/infoview.ts#L195

PR #284, opened 2021-12-03
Abbreviations `\/` and `\quot` for `⧸`

Vierkantor:
[Mathlib PR 10501](https://github.com/leanprover-community/mathlib/pull/10501) introduces `⧸` as an infix operator producing quotient types. I propose that we use `\/` and `\quot` as abbreviations for this operator.

I'm not quite sure what the best place would be to add these - I went for adding "quot" right after the alphabetically previous "quad", and "/" next.

* [Emacs PR](https://github.com/leanprover/lean-mode/pull/38)


src/abbreviation/abbreviations.json
```diff
@@ -962,6 +962,8 @@
     "cap": "∩",
     "qed": "∎",
     "quad": " ",
+    "quot": "⧸",
+    "/": "⧸",
```
eric-wieser, line 966: I wonder if it's worth having `bigsolidus` too since that's the Unicode name

Vierkantor: Sure, why not:
```suggestion
    "bigsolidus": "⧸",
    "/": "⧸",
```

gebner:
I've deployed a release, and cherry-picked the changes to vscode-lean4 and lean.nvim.

PR #301, opened 2022-05-18
fix: translate between utf16 code unit offsets (vscode) and codepoints (lean)

eric-wieser:
This testcase now draws the errors in the right place
```lean
example (𝒜𝒜𝒜𝒜𝒜𝒜𝒜𝒜𝒜𝒜𝒜𝒜 : nat) : ℕ := _ + _

example (AAAAAAAAAAAA : nat) : ℕ := _ + _
```

Also now working and previously broken:

* Hole positioning
* Infoview column numbers
* Autocomplete
* Goto definition
* Docstring hovers
* Try this (previously replaced the wrong text)
* `progressMessages` setting underlines, test-case:
  ```lean
  /-𝒜𝒜𝒜𝒜𝒜𝒜𝒜-/ #eval do tactic.sleep 1000
  /-𝒜𝒜𝒜𝒜𝒜𝒜𝒜-/ #eval do tactic.sleep 1000
  ```
Untested:

* `LeanWorkspaceSymbolProvider` (does the server even support this?)

### DIFF
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -2,6 +2,7 @@ import { CompletionItem, CompletionItemKind, CompletionItemProvider,
     MarkdownString, Position, Range, TextDocument, workspace } from 'vscode';
 import { Server } from './server';
 import { isInputCompletion } from './util';
+import { CodePointPosition } from './utils/utf16Position';
 
 const keywords = [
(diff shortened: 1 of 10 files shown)


gebner:
Thanks!

PR #305, opened 2022-06-09
Add support for listing symbols

eric-wieser:
With this input:
```lean
namespace ns

structure my_struct :=
(x : nat)

end ns

lemma is_eq : 1 = 1 := rfl

namespace ns

def f (n : nat) := 3 * n

end ns
```
This produces

image

The symbol choices are consistent with the choices in #306.

---

This depends on:
* https://github.com/leanprover/lean-client-js/pull/45
* #306


src/symbol.ts
```diff
@@ -0,0 +1,97 @@
...
+                    infos.push(curr);
+                }
+                stack.push(curr)
+            }
+
+            stack[stack.length-1].kind = this.toSymbolKind(item.kind);
+            stack[stack.length-1].detail = item.kind;
+        }
+
+        // An alternative approach that does not preserve order
```
gebner, line 61: Please remove the commented-out code (and add a comment above that order is preseved and why that is important).

src/symbol.ts
```diff
@@ -0,0 +1,97 @@
...
+import { CodePointPosition } from './utils/utf16Position';
+
+export class LeanDocumentSymbolProvider implements DocumentSymbolProvider {
+    server: Server;
+
+    constructor(server: Server) {
+        this.server = server;
+    }
+
+    toSymbolKind (s: string ) : SymbolKind {
```
gebner, line 14:
```suggestion
    toSymbolKind(s: string) : SymbolKind {
```

src/symbol.ts
eric-wieser: Removed and commented

eric-wieser:
Note this will need a release of `lean-client-js` to be made before it will pass CI

eric-wieser:
@gebner, mind taking another look now that all the dependencies are resolved?

PR #312, opened 2022-09-06
Proposal for an API based "Suggestions" detail in infoview

timlacroix:
As described in the title.
Some missing things on which I'd like help :
- clicking -> paste into editor. I imagine I should do something similar to 
```js
    text = text.replace(trythis.regexGM, (_, tactic) => {
        const command = encodeURI('command:_lean.pasteTacticSuggestion?' +
            JSON.stringify([m, tactic]));
        return `${trythis.magicWord}<a class="link" href="${command}" title="${tactic}">${tactic}</a>`
    });
```
But I'm not sure how to build $m$ in my case.

- debouncing queries to the API when editing the "prefix" field. (If we want to keep this field. Not sure it's super valuable)
- Suggestions status change (loading or not) doesn't seem perfect, refreshing suggestions requires a bit of cursor fiddling in the editor currently, and I think behaviour is weird on file change. Maybe I'm not handling widgets properly ?


gebner:
> clicking -> paste into editor. I imagine I should do something similar to

The `pasteTacticSuggestion` stuff replaces the current tactic, which is not what we want here.

What we actually want here is to insert a tactic before the current one.  But doing this properly is pretty hard in Lean 3, so I would suggest to just insert the suggestion on the line above.  This should do the trick:
```typescript
    post({ command: 'insert_text', text: `\n${suggestion}\n`, insert_type: 'relative'});
```

infoview/suggestions.tsx
```diff
@@ -0,0 +1,148 @@
...
+    useEvent(
+        SuggestionsErrorEvent,
+        (s) => {
+            setErrorMsg(s.error);
+        }, []
+    );
+
+    React.useEffect(() => {
+        setDoSuggest(true);
+    }, [prefix, props.goalState, props.widget])
```
gebner, line 107:
```suggestion
    }, [prefix, props.goalState])
```

This is the reason why it tries to update the suggestion on every cursor move.

React compares the dependency list using `Object.is`, i.e., reference equality.  The `props.widget` field is an object that is fetched every time you move the cursor, so it will be a different reference every time.

We don't need (or read) widgets anywhere, so I think you can just remove all occurrences of `widget`.

infoview/suggestions.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+    const inputProps = { style: { display: 'inline' }, type: 'text', size: 12, value: prefix };
+
+    return (
+        <div>
+            <details open={false}>
+                <summary className={'mv2'}>
+                    Suggestions
+                </summary>
+                <br/>
```
gebner, line 134:
```suggestion
```

Too much whitespace.

src/infoview.ts
```diff
@@ -388,4 +397,37 @@ export class InfoProvider implements Disposable {
             </body>
             </html>`
     }
+
+    private getSuggestions(reqId: number, goalState: string, prefix: string): void {
+        const api_key : string = workspace.getConfiguration('lean').get('suggestionAPIKey');
+        const url : string = workspace.getConfiguration('lean').get('suggestionURL');
```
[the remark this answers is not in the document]
timlacroix: A question mark symbol that you click on to "activate/deactivate" the suggestions detail which is then open by default ?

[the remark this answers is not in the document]
gebner: Yes, that sounds good.  Though I would default to closed by default.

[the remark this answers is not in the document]
timlacroix: suggestions are not activated by default, but if you click on the button, the detail is immediately open is what I meant

infoview/info.tsx
```diff
@@ -225,6 +210,9 @@ export function Info(props: InfoProps): JSX.Element {
                             </div>
                         </Details>
                 </div>
+                {doSuggest ? <div>
```
gebner, line 213:
```suggestion
                {(doSuggest && goalState) ? <div>
```

infoview/suggestions.tsx
```diff
@@ -0,0 +1,157 @@
...
+        },
+        [props.reqId]
+    );
+
+    React.useEffect(() => {
+        setStatus('loading');
+        setErrorMsg(undefined); // reset error msg
+    }, [props.reqId]);
+
+    return (
```
gebner, line 76:
```suggestion
    return goalState && (
```

src/infoview.ts
```diff
@@ -388,4 +397,45 @@ export class InfoProvider implements Disposable {
...
+        if (!url) {
+            void this.postMessage({
+                command: 'Suggestions',
+                results: {'error': 'lean.suggestionURL not set.'},
```
gebner, line 407:
```suggestion
                results: {'error': 'lean.suggestionURL not set, please ask on Zulip for URL and API key.'},
```

infoview/info.tsx
```diff
@@ -194,6 +178,7 @@ export function Info(props: InfoProps): JSX.Element {
                     <a className="link pointer mh2 dim" onClick={e => { e.preventDefault(); onPin(!isPinned)}} title={isPinned ? 'unpin' : 'pin'}>{isPinned ? <PinnedIcon/> : <PinIcon/>}</a>
                     <a className="link pointer mh2 dim" onClick={e => { e.preventDefault(); setPaused(!isPaused)}} title={isPaused ? 'continue updating' : 'pause updating'}>{isPaused ? <ContinueIcon/> : <PauseIcon/>}</a>
                     { !isPaused && <a className={'link pointer mh2 dim'} onClick={e => { e.preventDefault(); forceUpdate(); }} title="update"><RefreshIcon/></a> }
+                    <a className="link pointer mh2 dim" onClick={e => { e.preventDefault(); setDoSuggest(!doSuggest)}} title={doSuggest ? 'stop suggestions' : 'activate suggestions'}>{doSuggest ? <DoNotSuggestIcon/> : <DoSuggestIcon/>}</a>
```
gebner, line 181:
```suggestion
                    <a className="link pointer mh2 dim" onClick={e => { e.preventDefault(); setDoSuggest(!doSuggest)}} title={doSuggest ? 'stop ML suggestions' : 'activate ML suggestions'}>{doSuggest ? <DoNotSuggestIcon/> : <DoSuggestIcon/>}</a>
```
Not sure if this is the best name, but it distinguishes it from the other "suggestion" features we already have.

infoview/suggestions.tsx
```diff
@@ -0,0 +1,157 @@
...
+    ), [prefix, props.goalState, doSuggest])
+
+    const inputProps = { style: { display: 'inline' }, type: 'text', size: 12, value: prefix };
+
+    return (
+        <div>
+            <details open={true}>
+                <summary className={'mv2'}>
+                    Suggestions
+                </summary>
```
gebner, line 143:
```suggestion
```

infoview/suggestions.tsx
```diff
@@ -0,0 +1,157 @@
...
+                    ? <i className={'easeTransition mv2 ' + statusColTable.error}>{errorMsg}</i>
+                    : <div>
+                        <form className='ml2' onSubmit={(e) => e.preventDefault()}>
+                            <span>Tactic prefix: </span>
+                            <input {...inputProps} onChange={(e) => setPrefix(e.target.value)} />
+                        </form>
+                        <Suggestions widget={props.widget} goalState={props.goalState} reqId={suggReqId}/>
+                      </div>
+                }
+            </details>
```
gebner, line 154:
```suggestion
```

infoview/suggestions.tsx
```diff
@@ -0,0 +1,157 @@
...
+        <div>
+            <details open={true}>
+                <summary className={'mv2'}>
+                    Suggestions
+                </summary>
+                {errorMsg
+                    ? <i className={'easeTransition mv2 ' + statusColTable.error}>{errorMsg}</i>
+                    : <div>
+                        <form className='ml2' onSubmit={(e) => e.preventDefault()}>
+                            <span>Tactic prefix: </span>
```
gebner, line 148:
```suggestion
                            <span>Tactic suggestions with prefix: </span>
```

The double collapsing (with both the button and the details) is a bit weird.

gebner:
The mechanics are great now.  I think it's good enough to try out on real users now.

PR #319, opened 2022-11-17
fix(abbreviations.json): update norm symbol

fpvandoorn:
* Change the symbol associated to `\||` and `\fuzzy`
* `\parallel` gave the same symbol as `\||` before, so we don't change that one
* Remove `\shortparallel` which was the same as the previous `\||`
* Remove `\nshortparallel` and `\||n` which are the same as `\nparallel`
* Related to leanprover-community/mathlib#17575


src/abbreviation/abbreviations.json
```diff
@@ -909,7 +907,8 @@
     "||=": "⊫",
     "|||-": "⊪",
     "||n": "∦",
```
eric-wieser, line 909: I think we should remove this one and rename it to `paralleln` or `nparallel`

fpvandoorn: `nparallel` already exists for this symbol; I removed this one.

sgouezel:
bors r+
Thanks!

sgouezel:
(maybe there's no bors here...)

eric-wieser:
Nope, this needs a manual merge by @gebner (and of course a release for it to make any difference)

PR #321, opened 2022-11-23
New API suggestion icon

timlacroix:
By popular demand, since the question mark was confusing. 

image

image


alreadydone:
The new icons look good to me, thanks!

PR #328, opened 2023-02-01
Correct the symbol for perpendicular

eric-wieser:
Unicode has two separate symbols for "up tack" (ie `has_bot.bot`) and "perpendicular" (not yet used in mathlib). Having `\perp` produce the former is confusing.

### DIFF
--- a/src/abbreviation/abbreviations.json
+++ b/src/abbreviation/abbreviations.json
@@ -546,7 +546,7 @@
     "parallel": "∥",
     "pa": "▰",
     "pm": "±",
-    "perp": "⊥",
+    "perp": "⟂",
     "^perp": "ᗮ",
     "permil": "‰",
     "per": "⅌",


gebner:
Note that we already heavily use ⊥ in mathlib (and not ⟂).  For such a change I would like to see a consensus on zulip, and a commitment to a transition plan in both mathlib3 and mathlib4.

Because as soon as we release this change, people will rightfully complain that `\perp` no longer works.

gebner:
We don't use ⊥ just for bottom elements.  We also use the character for some orthogonality relations, like mutually singular measures.

eric-wieser:
I've created a [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Perpendicular.20notation/near/327158463). Thanks for finding the other uses!

eric-wieser:
Zulip consensus is unanimous; I don't think there's any lean 4 consideration here besides updating the abbreviations for that extension too; and I suspect we are already out of sync there.

PR #339, opened 2023-09-08
Update README to reflect that extension is for Lean 3

mhuisi:


gebner:
CI failure is unrelated.

